PR #4, opened 2017-12-27
Parse lists as equals-seq, same as we do for vectors

rafaeldff:
The use case was for checking a function that returned
the result of schema.core/explain.

Something like
```
(fact
  (s/explain {:a (s/enum :foo)}) => (match {:a '(s/enum :foo)}))
```

Felt weird to write
```
(fact
  (s/explain {:a (s/enum :foo)}) => (match {:a [s/enum :foo]}))
```
even though it matches in the same way.

An alternative would be to parse lists as `equals-value`, just like
primitives.

### DIFF
--- a/test/matcher_combinators/core_test.clj
+++ b/test/matcher_combinators/core_test.clj
@@ -265,18 +265,21 @@
     (#'core/match-any-order matchers [5] true)
     => [:mismatch (model/->Mismatch matchers [5])]))
 
-(fact "repro bug"
-  (core/match
-    (in-any-order :a
-                    [{:a {:id 1} :b 42}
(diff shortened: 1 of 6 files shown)


test/matcher_combinators/core_test.clj
```diff
@@ -264,3 +264,22 @@
     => [:mismatch (model/->Mismatch matchers [5])]
     (#'core/match-any-order matchers [5] true)
     => [:mismatch (model/->Mismatch matchers [5])]))
+
+(future-fact "reproducing select? bug"
```
rafaeldff: This future-fact refers to another bug, which will take a bit more thinking to fix.

PR #5, opened 2017-12-28
Allow (match) expressions inline within midje (provided) blocks

rafaeldff:
Support using match as any other midje checker:

```clj
(fact 
  (f [1 2 3]) => 1
  (provided
        (f (ch/match (m/equals-seq [int? int? int?]))) => 1))
```

### DIFF
--- a/src/matcher_combinators/midje.clj
+++ b/src/matcher_combinators/midje.clj
@@ -16,7 +16,7 @@
         true
         (checking/as-data-laden-falsehood {:notes [(printer/as-string result)]})))))
 
-(defmacro match [matcher-form]
+(defmacro ^{:midje/checker true} match [matcher-form]
   `(checkers.defining/as-checker
      (fn [actual#]
        (let [matcher# ~matcher-form]
--- a/test/matcher_combinators/midje_test.clj
+++ b/test/matcher_combinators/midje_test.clj
@@ -1,8 +1,9 @@
 (ns matcher-combinators.midje-test
-  (:require [midje.sweet :as midje :refer [fact facts =>]]
+  (:require [midje.sweet :as midje :refer [fact facts => falsey]]
             [matcher-combinators.midje :as ch]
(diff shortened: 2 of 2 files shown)


src/matcher_combinators/midje.clj
```diff
@@ -16,7 +16,7 @@
         true
         (checking/as-data-laden-falsehood {:notes [(printer/as-string result)]})))))
 
-(defmacro match [matcher-form]
+(defmacro ^{:midje/checker true} match [matcher-form]
   `(checkers.defining/as-checker
```
phillipm: `as-checker` adds `:midje/checker` metadata to forms, so I just realized that it might be possible to do 
```clojure
(checkers.defining/as-checker `(fn ...
```
instead of 
```clojure
`(checkers.defining/as-checker (fn ...
```
to address this issue instead of manually adding the metadata to the macro.

Not entirely sure if it will work but worth trying instead of doing `as-checker` and also adding the `:midje/checker` metadata

rafaeldff: I don't think that will work because I believe the metadata has to be affixed to the var, not the expanded expression, since midje will look for it in its own macros. I'll have another look tomorrow and check.

rafaeldff: Moving the syntax quote didn't work.

But I'm wondering if `match`really needs to be a macro. Can you remember the reason @phillipm?

The definition below seems to pass the tests.

```
(checkers.defining/defchecker match [matcher-form]
  (checkers.defining/checker [actual]
    (let [matcher matcher-form]
      (if (core/matcher? matcher)
        (check-match matcher actual)
        (ex-info "Input wasn't a matcher" {:input matcher})))))
```

rafaeldff: I'm merging this PR, but we can iterate on the checker in future pull requests.

PR #10, opened 2018-02-01
Show minimal diff for in-any-order mismatches

phillipm:
When `in-any-order` doesn't match, it shows a huge diff between the entire expected and actual arguments.

This PR addresses that by keeping the matcher ordering that leads to the most sub-elements being matched. With this info, when there isn't a full match, `in-any-order` can return the match results from the ordering that lead to the most matches.

| before | after |
| --------- | ---------- |
| before | after |

### DIFF
--- a/test/matcher_combinators/matchers_test.clj
+++ b/test/matcher_combinators/matchers_test.clj
@@ -1,10 +1,11 @@
 false(ns matcher-combinators.matchers-test
-  (:require [midje.sweet :as midje :refer [fact facts => falsey contains just anything]]
+  (:require [midje.sweet :as midje :refer [fact facts => falsey contains just anything future-fact has]]
             [matcher-combinators.midje :refer [match]]
+            [matcher-combinators.helpers :as helpers]
             [matcher-combinators.matchers :as m]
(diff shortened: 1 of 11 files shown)


test/matcher_combinators/matchers_test.clj
```diff
@@ -0,0 +1,70 @@
...
+
+(facts "show how input ordering affects diff size (when it ideally shouldn't)"
+  (fact "Given a particular input ordering, in-any-order shows the smallest diff"
+    (->> [{:a 2} {:b 2}]
+         (c/match (m/in-any-order [{:a 1} {:a 1 :b 2}]))
+         second
+         (map vals))
+    => (has every? one-mismatch?))
+
+  (future-fact "in-any-order minimization doesn't find the match ordering that
```
phillipm: This test points to the cases that this PR doesn't address: true 'minimal' matching in the sense of traversing the nested match results to count all mismatches instead of just finding the matcher ordering that leads to the least top-level mismatches

phillipm:
@rafaeldff I cleaned this up and it is ready to be reviewed

PR #20, opened 2018-02-18
Rename matchers and implement set matchers

phillipm:
We currently have unique matcher names for maps, sequences, and values.
This potentially introduces overhead for the user to remember the unique names in every context (I say potentially because I'm undecided if it is better to overload names or have unique names for every specific matcher).

This PR explores the idea of renaming
 - `equals-map`, `equals-seq`, and `equals-value` to `equals` 
 - `contains-map` and `subset` to `embeds`
- `subseq` to `prefix-seq`
and dispatch to the correct matcher logic based on the shape of the `expected` argument

Also introduced are matchers for sets
 - `equals` and `embeds` work for sets
 - `set-equals` and `set-embeds` allow one to use the same submatcher more than once by accepting a sequence instead of a set as the `expected` argument. That is `(match (equals #{odd? odd?}) #{1 3})` will fail because `#{odd? odd?}` evaluates to `#{odd?}`. So one can use `(match (set-equals [odd? odd?]) #{1 3})`

### DIFF
--- a/src/matcher_combinators/matchers.clj
+++ b/src/matcher_combinators/matchers.clj
@@ -7,14 +7,16 @@
   [expected]
   (cond
     (sequential? expected) (core/->EqualsSeq expected)
-    (set? expected)        (core/->EqualsSet expected)
+    (set? expected)        (core/->EqualsSet expected false)
     (map? expected)        (core/->EqualsMap expected)
     :else                  (core/->Value expected)))
 
 (defn equals-set
-  ""
+  "Matches a set in the way `(equals some-set)` would, but accepts sequences as
+  the expected matcher argument, allowing one to use matchers with the same
+  submatcher appearing more than once."
   [expected]
-  (core/->EqualsSet expected))
+  (core/->EqualsSet expected true))
 
 (defn contains
   "Matcher that will match when the map contains some of the same key/values as
@@ -22,14 +24,16 @@
   [expected]
   (cond
     (sequential? expected) (core/->ContainsSeq expected)
-    (set? expected)        (core/->ContainsSet expected)
+    (set? expected)        (core/->ContainsSet expected false)
     (map? expected)        (core/->ContainsMap expected)
     :else                  (core/->InvalidType expected "contains" "seq, set, map")))
 
 (defn contains-set
-  ""
+  "Matches a set in the way `(contains some-set)` would, but accepts sequences
+  as the expected matcher argument, allowing one to use matchers with the same
+  submatcher appearing more than once."
   [expected]
-  (core/->ContainsSet expected))
+  (core/->ContainsSet expected true))
 
 (defn in-any-order
   "Matcher that will match when the given a list that is the same as the
--- a/src/matcher_combinators/matchers.clj
+++ b/src/matcher_combinators/matchers.clj
@@ -1,31 +1,35 @@
 (ns matcher-combinators.matchers
   (:require [matcher-combinators.core :as core]))
 
-(defn equals-value
+(defn equals
   "Matcher that will match when the given value is exactly the same as the
   `expected`."
   [expected]
-  (core/->Value expected))
+  (cond
+    (sequential? expected) (core/->EqualsSeq expected)
+    (set? expected)        (core/->EqualsSet expected)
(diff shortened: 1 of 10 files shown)


src/matcher_combinators/matchers.clj
```diff
@@ -1,31 +1,39 @@
...
+
+(defn equals-set
+  "Matches a set in the way `(equals some-set)` would, but accepts sequences as
+  the expected matcher argument, allowing one to use matchers with the same
+  submatcher appearing more than once."
+  [expected]
+  (core/->EqualsSet expected true))
 
-(defn contains-map
+(defn embeds
```
phillipm: I chose `embeds` because it won't conflict with midje's `contains` when people do `[midje.sweet :refer :all]`

src/matcher_combinators/matchers.clj
```diff
@@ -1,31 +1,39 @@
...
+(defn equals
   "Matcher that will match when the given value is exactly the same as the
   `expected`."
   [expected]
-  (core/->Value expected))
+  (cond
+    (sequential? expected) (core/->EqualsSeq expected)
+    (set? expected)        (core/->EqualsSet expected false)
+    (map? expected)        (core/->EqualsMap expected)
+    :else                  (core/->Value expected)))
```
phillipm: is there a reason to use multimethods instead of this type of cond'ing to do function dispatch based on type?

src/matcher_combinators/matchers.clj
```diff
@@ -40,18 +48,10 @@
...
-(defn sublist
+(defn prefix-seq
   "Matcher that will match when provided a (ordered) prefix of the `expected`
   list.
 
-  Similar to Midje's `(contains expected)`"
-  [expected]
-  (core/->SubSeq expected))
-
-(defn subset
```
phillipm: matching an unordered subset of elements in a sequence used to be `subset`. It is now called `embeds`, does that seem like a good name for it?

src/matcher_combinators/parser.clj
```diff
@@ -38,6 +38,6 @@
                BigInt
                Character)
 
-(mimic-matcher matchers/contains-map IPersistentMap)
-(mimic-matcher matchers/equals-seq IPersistentVector)
-(mimic-matcher matchers/equals-seq IPersistentList)
+(mimic-matcher matchers/embeds IPersistentMap)
+(mimic-matcher matchers/equals IPersistentVector)
+(mimic-matcher matchers/equals IPersistentList)
```
phillipm: I might try to direct calls to the record creation functions like `->Value` instead of letting `equals` do it. Leaving it this way because I also might try to add some indirect to allow for an `match-exactly` that changes the default for maps to be `equals` instead of `embeds`.

src/matcher_combinators/matchers.clj
```diff
@@ -40,18 +48,10 @@
   ([select-fn expected]
    (core/->SelectingInAnyOrder select-fn expected)))
 
-(defn sublist
+(defn prefix-seq
```
rafaeldff: Why not just `prefix`?

src/matcher_combinators/matchers.clj
rafaeldff: I think `cond` seems fine for this purpose.

src/matcher_combinators/matchers.clj
```diff
@@ -1,31 +1,39 @@
...
+  (cond
+    (sequential? expected) (core/->EmbedsSeq expected)
+    (set? expected)        (core/->EmbedsSet expected false)
+    (map? expected)        (core/->EmbedsMap expected)
+    :else                  (core/->InvalidType expected "embeds" "seq, set, map")))
+
+(defn embeds-set
+  "Matches a set in the way `(embeds some-set)` would, but accepts sequences
+  as the expected matcher argument, allowing one to use matchers with the same
+  submatcher appearing more than once."
```
rafaeldff: Do you have real use cases for `embeds-set`? At first sight, it seems to be more specific and a bit harder to understand than the rest of the matchers. 

BTW, is it `embeds-set` or `set-embeds`?

PR #33, opened 2018-04-30
Add support for java.lang.Integer and java.lang.Float

plexus:
Currently when comparing numbers of these types an exception will be raised,
since they don't implement the Matcher protocol.

### DIFF
--- a/src/matcher_combinators/parser.clj
+++ b/src/matcher_combinators/parser.clj
@@ -22,7 +22,9 @@
 
 (mimic-matcher matchers/equals
                nil
+               Integer
                Long
+               Float
                Double
                String
                Symbol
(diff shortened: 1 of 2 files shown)


phillipm:
Thanks for the fix!

Can you add some test coverage by including the following generators in the [`gen-scalar`](https://github.com/nubank/matcher-combinators/blob/53183b11f54802e14a16b12ea8ffd48463a9f3f4/test/matcher_combinators/parser_test.clj#L18) list?

```clojure
(def gen-java-integer
  (gen/fmap #(Integer. %) gen/int))

(def gen-float
  (gen/fmap #(float %) gen/double))
```

plexus:
Of course! Thanks for showing me the way. How does this look?

phillipm:
to follow up, these changes are now available on clojars in [`0.2.5`](https://github.com/nubank/matcher-combinators/blob/master/CHANGELOG.md#025)

PR #44, opened 2018-08-31
Add regex matcher

phillipm:
In Midje checkers you can write a regex and it uses it as a checking predicate.
The same would be nice for matcher-combinators

For example:
```clojure
(fact "treat regex as predicate in match"
  {:one "1"} => (match {:one #"1"}))
```

example mismatch output

```clojure
(fact "when a regex doesn't match"
  {:one "1"} => (ch/match {:one #"2"}))
```
results in:
```
FAIL "treat regex as predicate in match" at (midje_test.clj:220)
Actual result:
{:one "1"}
Checking function: (ch/match {:one #"2"})
        [:mismatch {:one (mismatch #"2" "1")}]
```

and

```clojure
(fact "trying to apply regex to wrong type"
  {:one 1} => (ch/match {:one #"2"}))
```
results in:
```
Actual result:
{:one 1}
Checking function: (ch/match {:one #"2"})
        [:mismatch
 {:one
  (invalid-matcher-input
   "regex #\"2\" can't match 'expected' argument of type: class java.lang.Long"
   "provided: 1")}]
```

### DIFF
--- a/src/matcher_combinators/core.clj
+++ b/src/matcher_combinators/core.clj
@@ -49,6 +49,22 @@
      :else
      nil)))
 
+(defn- regex? [value] (instance? java.util.regex.Pattern value))
+(defrecord Regex [expected]
+  Matcher
+  (match [_this actual]
+   (if-let [issue (validate-input expected actual regex? (constantly true) 'regex "java.util.regex.Pattern")]
+     issue
+     (try
+       (if (re-find expected actual)
+         [:match actual]
+         [:mismatch (model/->Mismatch expected actual)])
+       (catch ClassCastException ex
+         [:mismatch (model/->InvalidMatcherType
+                      (str "provided: " actual)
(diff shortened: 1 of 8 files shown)


src/matcher_combinators/core.clj
```diff
@@ -49,6 +49,22 @@
...
      nil)))
 
+(defn- regex? [value] (instance? java.util.regex.Pattern value))
+(defrecord Regex [expected]
+  Matcher
+  (match [_this actual]
+   (if-let [issue (validate-input expected actual regex? (constantly true) 'regex "java.util.regex.Pattern")]
+     issue
+     (try
+       (if (re-find expected actual)
```
claynon: have you considered using `re-matches`? I don't know if there is any difference in performance
(this more of a bike sheeding question)

claynon: 🤔 is the output of these functions used when they match?
Because when they don't match `re-find` just returns `nil` right?

PR #45, opened 2018-09-04
Add regex matcher

caioaao:
I couldn't test the mismatch case for some reason. What I get:

```
Diffs: in [1 :expected] expected #"^ref", was #"^ref"
```

Trying to use `exactly` also fails


phillipm:
Crazy, this is an issue I was looking at the other day because Claynon also ran into this. I started a PR on it https://github.com/nubank/matcher-combinators/pull/44
That PR needs a few more tests that exercise matching and mismatching regexes, to make sure I didn't mess up the matching logic, so I was waiting until then to merge it.

caioaao:
:man_facepalming: github search sucks :\ sorry and thanks! Also, just as another data point, I'm already using this implementation on a project and looks like it's working well :)

Issue #47, opened 2018-09-14
Embeds for `seq` does not work like `(contains :in-any-order :gaps-ok)`


oporkka:
I am using matcher-combinators 0.3.1, and comparing Midje's
```
  (fact "contains :in-any-order :gaps-ok"
    [1 2 3 4 5 6] => (contains [1 5]
                               :in-any-order
                               :gaps-ok))
```
to
```
  (fact "m/embeds"
    [1 2 3 4 5 6] => (match (m/embeds [1 5])))
```
The version with `contains :in-any-order :gaps-ok` passes, but the latter one with `embeds` does not pass.  The README says for `embeds` that:

- sequence: order-agnostic matcher that will match when provided a subset of the expected sequence. Similar to midje's `(contains expected :in-any-order :gaps-ok)`

It seems that there's either a mismatch in the implementation or the documentation.

phillipm:
Indeed, very strange! Thanks for reporting, I'm going to try to get this sorted

phillipm:
Resolved in https://github.com/nubank/matcher-combinators/pull/48 and deployed as `0.3.2`. Cheers!

PR #48, opened 2018-09-14
Fix for seq embeds matching due to not all permutations being computed

phillipm:
Fix for https://github.com/nubank/matcher-combinators/issues/47

Was caused because the permutation helper I had defined wasn't computing all permutations, only computing permutations created by rotating the sequence.

Calculating all permutations might be heavy handed, but prefer to refactor it once it proves to be a bottleneck

### DIFF
--- a/test/matcher_combinators/core_test.clj
+++ b/test/matcher_combinators/core_test.clj
@@ -290,7 +290,6 @@
     => (just [:mismatch (just [5 (just (model/->Missing anything))]
                               :in-any-order)])))
 
-
 (tabular
   (fact "Providing seq/map matcher with incorrect input leads to automatic mismatch"
     (core/match (?matcher 1) 1)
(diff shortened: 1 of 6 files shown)


test/matcher_combinators/core_test.clj
```diff
@@ -307,6 +306,12 @@
 (def pred-set #{(pred-matcher odd?) (pred-matcher pos?)})
 (def pred-seq [(pred-matcher odd?) (pred-matcher pos?)])
 
+(def short-equals-seq (map equals [1 3]))
+
+(fact "embeds for sequences"
+  (core/match (embeds short-equals-seq) [3 4 1]) => (just [:match (just [3 4 1])])
+  (core/match (embeds short-equals-seq) [3 4 1 5]) => (just [:match (just [3 4 1 5])]))
```
phillipm: this case wasn't working before because `1` and `3` were never going to be next to each other in the permutations, so the way subset matching was done, this was causing issues

PR #49, opened 2018-09-18
[clojure.test] Use regular :fail instead of custom :mismatch message

plexus:
Hi @phillipm, we ran into issues with CIDER and matcher-combinators. I know a fix has been suggested [here](https://github.com/clojure-emacs/cider/issues/2215#issuecomment-368269371), but that essentially involves patching/extending CIDER and every other tool that does its own test running. I was wondering if it wouldn't be possible to fix this on the matcher-combinator side so it's compatible with any clojure.test based tool. 

In case of a mismatch we emit a custom map with {:type :mismatch} in order to do
custom rendering. The problem with this approach is that other tools and
reporters (CIDER, Cursive, eftest, ...) all need to be aware of :mismatch.

Instead emit a standard {:type :fail} method, but extend
`clojure.core/print-method` to still get the pretty printing of mismatch
information. This does mean we are forced in the straightjacket of "Expected:
...\nActual: ...".

This PR is just for demonstration purposes. I'm curious what you think of this approach.

Here is how it looks

2018-09-18-115923_584x118_scrot

Here is how it comes out in CIDER

2018-09-18-115726_806x442_scrot

A bit unfortunate that they also do their own rendering, so the `print-method` implementation doesn't get picked up, but it's usable.

### DIFF
--- a/src/matcher_combinators/test.clj
+++ b/src/matcher_combinators/test.clj
@@ -25,35 +25,28 @@
        stacktrace-file-and-line
        (merge report)))
 
+(defmethod clojure.core/print-method ::mismatch [result out]
+  (binding [*out* out]
+    (println)
+    (printer/pretty-print result)))
+
 (defmethod clojure.test/assert-expr 'match? [msg form]
   `(let [[matcher# actual#] (list ~@(rest form))]
      (if (core/matcher? matcher#)
        (let [result# (core/match matcher# actual#)]
          (clojure.test/do-report
-           (if (core/match? result#)
-             {:type     :pass
-              :message  ~msg
-              :expected '~form
-              :actual   (list 'match? matcher# actual#)}
-             (with-file+line-info
-               {:type     :matcher-combinators/mismatch
-                :message  ~msg
-                :expected '~form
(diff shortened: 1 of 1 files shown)


phillipm:
Thanks for looking into this! 
I've had my head down in some other stuff and haven't be able to dive into what is going on here. I'm hoping to take a look soon :)

I did try running a failing `clojure.test` test via `clojure.test/run-tests` and I didn't get highlighted output that is shown in your first screenshot.

```
(is (match? (m/equals [1 odd?]) [2 3]))
```

```
FAIL in (basic-matching) (test_test.clj:25)

expected: (matcher-combinators.test/match? ...)
  actual: [#matcher_combinators.model.Mismatch {:actual 2 :expected 1} 3]
```

thumbnail:
Hate to bump; but is there any progress on this? It's rather blocking to have broken test reporting.

philomates:
Thanks for the bump @thumbnail I had forgotten about this.
I merged this (with some updates to deal with changes since the PR was opened)
I'm travelling and hence away from the computer that I use for deploys. So I'll release this change on monday.

thumbnail:
Thanks for the quick follow up! Awesome work.

PR #50, opened 2018-09-19
Plexus/try out kaocha

plexus:
Hi @phillipm, I've pushed kaocha-midje to Clojars and updated your branch. This is just to demo kaocha, I don't expect you to merge this.

Some things to note

- it just uses leiningen, you don't need deps.edn to use Kaocha
- the `tests.edn` doesn't need to contain much, just a test suite of type `:kaocha.type/midje`
- Kaocha uses tools namespace 0.3.0-alpha, which clashes with the version that midje uses... I've excluded the one from midje for now, but I'm beginning to think I might need to vendor/mr-andersonize this library inside kaocha
- `--watch` now uses tools.namespace as you suggested, and seems to work ok now
- test are loaded and being run, but some of them fail according to kaocha, and the output is far from helpful, so there's still some work there.
- I've given kaocha-midje its proper home at https://github.com/lambdaisland/kaocha-midje, even though it's clearly still very alpha

Still I'll count this as a minor success, it's loading, randomizing, running, and reloading tests. So far so good.


bin/kaocha
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+lein kaocha "$@"
```
plexus: As per the docs I recommend every project to have a `bin/kaocha` wrapper, so it can be used consistently regardless of which tool you're using.

project.clj
```diff
@@ -12,7 +12,12 @@
   :dependencies [[org.clojure/clojure "1.9.0"]
                  [colorize "0.1.1" :exclusions [org.clojure/clojure]]
                  [org.clojure/math.combinatorics "0.1.4"]
-                 [midje "1.9.2" :exclusions [org.clojure/clojure]]]
+                 [midje "1.9.2" :exclusions [org.clojure/clojure
+                                             org.clojure/tools.namespace]]]
```
plexus: Without this kaocha's watch code fails, because it relies on the alpha.

project.clj
```diff
@@ -12,7 +12,12 @@
...
 
   :profiles {:dev {:dependencies [[org.clojure/test.check "0.9.0"]]}
-             :1.8 {:dependencies [[org.clojure/clojure "1.8.0"]]}})
+             :1.8 {:dependencies [[org.clojure/clojure "1.8.0"]]}
+             :kaocha {:dependencies [[lambdaisland/kaocha "0.0-153"]
+                                     [lambdaisland/kaocha-midje "0.0-5"]]}}
+
+  :aliases {"kaocha" ["with-profile" "+kaocha" "run" "-m" "kaocha.runner"]})
```
plexus: This could also just be in the wrapper script, but this way you can also do `lein kaocha --watch` for instance.

phillipm:
very cool! I'll give this a try soon.

I've bumped to the latest `tools.namespace` alpha in Midje `1.9.3-alpha2`

One thing I'm curious about:
Do you know why `lein` is so much faster than `deps.edn` for running `kaocha`? It is around double a fast, if I recall correctly

plexus:
I'll be looking at kaocha-midje again before the end of the month, I might ping you again then. Meanwhile closing this as it was just a demo anyway.

PR #64, opened 2019-03-10
Clojurescript support

philomates:
extends matcher-combinators to work with clojurescript.

I've never done clojurescript before and I have no idea what I'm doing, so if anyone has suggestions with regards to cleaner approaches, I'd be very grateful to hear them.


src/cljc/matcher_combinators/parser.cljc
```diff
@@ -0,0 +1,132 @@
...
+  #?(:clj
+     (:import [clojure.lang Keyword Symbol Ratio BigInt IPersistentMap
+               IPersistentVector IPersistentList IPersistentSet
+               LazySeq Repeat Cons Var]
+              [java.util UUID Date]
+              [java.util.regex Pattern]
+              [java.time LocalDate LocalDateTime LocalTime YearMonth])))
+
+#?(:cljs
+(extend-protocol
```
philomates: not really sure why the `mimic-matcher` macro doesn't work for cljs

PR #66, opened 2019-03-24
Implement `absent` matcher for maps

philomates:
currently
```clojure
;; matches
(is
  (match? {:a number? :b nil}
          {:a 1 :b nil}))
;; mismatches with `:b should have been nil`
(is
  (match? {:a number? :b nil}
          {:a 1}))
```
which I like because I haven't fully embraced clojure's nil-punning :man_shrugging: 

but in many cases it is nice to not have to write an additional test checking the absence of keys in a map.

Hence, this PR introduces the `absent` matcher to assert that a map doesn't have an entry for a key.

```clojure
;; matches
(is
  (match? {:a number? :b absent}
          {:a 1}))
;; doesn't match
(is
  (match? {:a number? :b absent}
          {:a 1 :b nil}))
```


src/cljc/matcher_combinators/matchers.cljc
```diff
@@ -54,3 +54,7 @@
   "Matcher that will match when given value matches the `expected` regular expression."
   [expected]
   (core/->Regex expected))
+
+(def absent
```
lucascs: should it be `absent?` then?

test/clj/matcher_combinators/core_test.clj
```diff
@@ -373,6 +373,44 @@
...
+              ::result/value  {:a 42}
+              ::result/weight 0})
+
+    (core/match (?matcher {:a (equals 42)
+                           :b absent})
+      {:a 42
+       :b 43})
+    => (just {::result/type   :mismatch
+              ::result/value  (just {:a 42
+                                     :b (just {:actual 43})})
```
lucascs: this error doesn't say it should be missing...

I'd like to see a print of a failure message on the terminal.

src/cljc/matcher_combinators/matchers.cljc
philomates: the other matchers, like `embeds` and `equals` don't have the `?` so I figured I'd keep the same pattern

test/clj/matcher_combinators/core_test.clj
philomates: The mismatch data-structures have pretty printers that turn these raw maps into output, so you don't really see the output in tests.

https://github.com/nubank/matcher-combinators/pull/66/commits/e6534d5898ff3583ed4b939190649da53c11c4bd is an example where you can tweak the values and rerun the tests to see the failure messages

for 
```clojure
(fact   {:a {:aa 11} :b 1} => (match {:a {:bb m/absent} :b m/absent}))
```
you get
```
FAIL map absent at (midje_test.clj:41)
Actual result:
{:a {:aa 11} :b 1}
Checking function: (match {:a {:bb m/absent}, :b m/absent})
        [:mismatch {:a {:aa 11}, :b (unexpected 1)}]
```

and 

```clojure
(fact "absent used in wrong context"
  [1] =not=> (match [m/absent]))
```
you get 
```
FAIL absent used in wrong context at (midje_test.clj:44)
Actual result:
[1]
Checking function: (match [m/absent])
        [:mismatch
 [(invalid-matcher-context
   "`absent` matcher should only be used as the value in a map")]]
```

PR #67, opened 2019-03-25
Match two records by type using =

mpivaa:
Actually, matching two records with different types and same values will result in a false positive `match`.


src/cljc/matcher_combinators/core.cljc
```diff
@@ -143,6 +143,16 @@
...
       (compare-maps expected actual model/->Unexpected false))))
 
+(defrecord EqualsRecord [expected]
+  Matcher
+  (match [_this actual]
+    (if-let [issue (validate-input expected actual map? 'equals "record")]
+      issue
+      (cond
+        (and (record? expected)
+             (record? actual)) (match (->Value expected) actual)
```
philomates: Mismatches for the same record type with different values doesn't show nice diffs. 
```
        [:mismatch (mismatch {:x 1, :y 2} {:x 2, :y 2})]
```
could be
```
        [:mismatch {:x (mismatch 1 2), :y 2}]
```
for 
```
(fact (->Point 2 2) => (match (m/equals (->Point 1 2))))
```

philomates: Also, I wonder if it is better to make is such that if the `expected` is a record then the match fails if the `actual` isn't.

philomates: So maybe something along the lines of 

```clojure
(defn- match-record [expected actual]
  (if (and (record? actual)
           (= (type expected) (type actual)))
    (match (->EqualsMap expected) actual)
    {::result/type   :mismatch
     ::result/value  (model/->RecordMismatch expected actual)
     ::result/weight 1}))

(defrecord EqualsRecord [expected]
  Matcher
  (match [_this actual]
    (if-let [issue (validate-input expected actual map? 'equals "record")]
      issue
      (cond
        (record? expected)
        (match-record expected actual)

        :else
        (match (->EqualsMap expected) actual)))))
```

test/clj/matcher_combinators/matchers_test.clj
```diff
@@ -148,3 +148,32 @@
                 ::result/value {:actual   b
                                 :expected a}
                 ::result/weight 1}))))
+
+(defrecord RecordA [val])
+(defrecord RecordB [val])
+
+(facts "records"
```
philomates: would be nice to add some tests in `midje-test` namespace just to see how this change plays out in practice

src/cljc/matcher_combinators/core.cljc
```diff
@@ -143,6 +143,16 @@
       issue
       (compare-maps expected actual model/->Unexpected false))))
 
+(defrecord EqualsRecord [expected]
```
philomates: an `EmbedsRecord` probably makes sense as well

src/cljc/matcher_combinators/core.cljc
```diff
@@ -143,6 +143,16 @@
...
+  Matcher
+  (match [_this actual]
+    (if-let [issue (validate-input expected actual map? 'equals "record")]
```
philomates: I think you should use the other arity of `validate-input` and do `(validate-input expected actual record? map? 'equals "record")`

src/cljc/matcher_combinators/core.cljc
mpivaa: ☝️ this would be a breaking change, but makes sense.

src/cljc/matcher_combinators/core.cljc
mpivaa: Seems like records always consider inexistent properties as `nil`, so it will behave like `equals`

mpivaa:
@philomates can we merge this?

philomates:
Hey @mpivaa sorry for the slow follow-up. I finally dived into this again and ended up making some tweaks described below. Let me know what you think and we can deploy it

Yeah, I guess the issue with the parser / `mimic-matcher` where we want to make the default for records be `equals` is due to the fact that records implement both `IPersistentMap` and `IRecord`, so we can't have the parser default to being `equals` for records but `embeds` for `maps`. In the face of this I think we should make the implementation of `embeds` for records point to `equals`. Ideally throwing an error when you use `embeds` on a record would happen, because as you mentioned, records can't be partial so taking the embeds of them doesn't really make sense.

I decided to go forward with making the breaking change mentioned in https://github.com/nubank/matcher-combinators/pull/67#discussion_r269323939 where if you expect a record but provide a map it fails. I think this is a good mix of explicitness and dynamism because you can assert class info for records but also compare them to maps if needed. What do you think?

src/cljc/matcher_combinators/matchers.cljc
```diff
@@ -1,15 +1,21 @@
 (ns matcher-combinators.matchers
-  (:require [matcher-combinators.core :as core]))
+  (:require [clojure.string :as string]
+            [matcher-combinators.core :as core]))
+
+(defn- non-internal-record? [v]
+  (and (record? v)
+       (not (string/starts-with? (-> v type str) "class matcher_combinators.core"))))
```
philomates: needed so we don't match against things like the `Absent` record in `core`

mpivaa:
Looks good! :shipit:

PR #68, opened 2019-03-27
Add declare to stop linters from erasing require

caioaao:
As it's not being directly used by the code that requires it, linters keep complaining of unused import when requiring `matcher-combinators.test`.

Now clients can write `[matcher-combinators.test :refer [match?]]` and stop linters from deleting this line. And, as it's not declared outside from clojure test's checks, so it still works as intended.


philomates:
should be up on clojars

PR #69, opened 2019-03-27
add `value` matcher

philomates:
I was trying to compare references to prismatic schemas using matcher-combinators and running into issues. This is because schemas are in the end maps, but they contain predicate functions. When you try to use matcher-combinators with these predicate functions, they actually get applied, resulting in undesirable behavior.

Thus I'm introducing a way to force using `=` to compare maps via the `value` matcher

### DIFF
--- a/src/cljc/matcher_combinators/matchers.cljc
+++ b/src/cljc/matcher_combinators/matchers.cljc
@@ -1,6 +1,11 @@
 (ns matcher-combinators.matchers
   (:require [matcher-combinators.core :as core]))
 
+(defn value
+  "Matcher that uses underlying language-level equality"
+  [expected]
+  (core/->Value expected))
+
 (defn equals
   "Matcher that will match when the given value is exactly the same as the
   `expected`."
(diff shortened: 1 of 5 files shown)


src/cljc/matcher_combinators/matchers.cljc
```diff
@@ -1,6 +1,11 @@
 (ns matcher-combinators.matchers
   (:require [matcher-combinators.core :as core]))
 
+(defn value
```
philomates: is there a better name? perhaps `reference` or `exact`?
Will this cause confusion because it is quite similar to `equals`?
Perhaps it is better to introduce map/seq specific value matcher like `map-reference` or something

philomates:
An alternative would be to load the following code every time you want to compare prismatic schemas, which still isn't ideal because it would match schemas loosely

```clojure
(ns my-project.utils.matcher-combinators-parsers
  (:require [matcher-combinators.core :as core]
            [matcher-combinators.parser :as parser])
  (:import [schema.core ConditionalSchema Predicate NamedSchema]))

;; NOTE: once matcher-combinators implements recursive strict matching on
;; nested data-structures, this should be updated to not point to ->Value
(parser/mimic-matcher (fn [x] (core/->Value x))
                      ConditionalSchema
                      Predicate
                      NamedSchema)
```

philomates:
For the this special case I think it is better to just do `(match {:schema #(= % some-prismatic-schema)})` and avoid introducing a new matcher that is almost the same as `equals`

PR #70, opened 2019-03-29
Treat nil value in map as expecting the key/value to be absent

philomates:
an alternative approach to https://github.com/nubank/matcher-combinators/pull/66

currently
```clojure
;; matches
(is
  (match? {:a number? :b nil}
          {:a 1 :b nil}))
;; mismatches with `:b should have been nil`
(is
  (match? {:a number? :b nil}
          {:a 1}))
```
which can be useful because `nil` is a base value and usually in matcher-combinators base values match using `=`.

but in many cases it is nice to not have to write an additional test checking the absence of keys in a map.

Hence, this PR changes the behavior of `nil` in the value position of a map to assert that the key isn't present in the map

```clojure
;; matches because `:b` is missing from the actual
(is
  (match? {:a number? :b nil}
          {:a 1}))
;; doesn't match because `:b` is present
(is
  (match? {:a number? :b nil}
          {:a 1 :b nil}))
```

if you'd like to assert that the value at a key is `nil` you should use the predicate `nil?`

```clojure
;; matching a nil value
(is
  (match? {:a number? :b nil?}
          {:a 1 :b nil}))
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,19 @@
 # Change Log
 All notable changes to this project will be documented in this file. This change log follows the conventions of [keepachangelog.com](http://keepachangelog.com/).
 
+## [1.0.0]
+- Using `nil` as the matcher for value at a key in a map now only matches if the key is absent.
+
+For example:
+```clojure
+(testing "will match because `:b` points to `nil` in the matcher and the `actual` doesn't have `:b`"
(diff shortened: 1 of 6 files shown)


philomates:
went with the more explicit https://github.com/nubank/matcher-combinators/pull/66 approach

Issue #63, opened 2019-02-07
Match `embeds` are very slow in this specific case


RafaeLeal:
Hi!

For some reason, this test is very slow.. almost 1 minute.

```clojure
(fact "something is slow"
      [{:a {:a "a",
            :b "b",
            :c "c",
            :d "d",
            :e "e",
            :f "f"}}
       {:a {:a "a",
            :b "b",
            :c "c",
            :d "d",
            :e "e",
            :f "f"}}
       {:a {:a   "foo",
            :b ["-c" "some-string"]}}
       {:a {:a   "foo",
            :b ["-c" "some-string-2"]}}
       {:a {:a   "foo",
            :b ["-c" "some-string-3"]}}
       {:a {:a   "foo",
            :b ["-c" "some-string-4"]}}
       {:a {:a   "foo",
            :b ["-c" "some-string-5"]}}
       {:a {:a   "foo",
            :b ["-c" "some-string-6"]}}
       {:a {:a   "foo",
            :b ["-c" "some-string-7"]}}
       {:a {:a   "foo",
            :b ["-c" "some-string-8"],
            :c    "good"}}] => (match (embeds [{:a {:a "foo" :b ["-c" "some-string-5"]}}])))
```

lein midje matchers-slow  51.11s user 1.54s system 108% cpu 48.531 total
lein midje matchers-slow  49.41s user 1.59s system 107% cpu 47.334 total
lein midje matchers-slow  52.27s user 1.70s system 107% cpu 50.197 total

while running with
```clojure
(match [anything
              anything
              anything
              anything
              anything
              anything
              {:a {:a "foo" :b ["-c" "some-string-5"]}}
              anything
              anything
              anything])
```
lein midje matchers-slow  15.62s user 1.22s system 122% cpu 13.777 total

philomates:
Hey @RafaeLeal thanks for reporting. When implementing some of the matchers I decided to use some algorithms with really bad BigO factors; deciding to get more comprehensive matching at the cost of runtime speed.
My intention was to revisit the decisions if anyone noticed things getting slow, which it appears you've noticed.

I'll try to take a look at this soon, but probably won't get a chance in the next week or two

philomates:
Thanks for reporting @RafaeLeal. I'm pretty sure I've fixed it in https://github.com/nubank/matcher-combinators/pull/72 which has been deployed as `0.8.2`

PR #71, opened 2019-04-02
[BUG] Test that takes a huge amount of time to complete

lucascs:
For some reason this test takes forever to complete (it's supposed to fail)


lucascs:
image
build failed by timeout

philomates:
Looks the same as https://github.com/nubank/matcher-combinators/issues/63

It is most likely due to the fact that we permute over both the matchers and the elements in the list to get the smallest mismatch: https://github.com/nubank/matcher-combinators/pull/10/files#diff-66c3cf76ecc4414d444248dce959d7a8R156

I'll have to dive deeper to see if there is a way to improve this and maintain the nice minimal mismatch. I'm a bit overloaded preparing for a talk but hopefully I can look at this next week.

philomates:
addressed in https://github.com/nubank/matcher-combinators/pull/72 and https://github.com/nubank/matcher-combinators/pull/87

PR #76, opened 2019-04-24
Create a timeout test case for in-any-order

comigor:
After https://github.com/nubank/matcher-combinators/pull/72 (`0.8.2`), one of our test cases using `in-any-order` timeouted. Before `0.8.2`, it worked. The test case was comparing two lists of 50 items each using `in-any-order`.

It seems `in-any-order` process time grows exponentially by the number of items it's comparing. By 10 items it'll definitely timeout any CI.

I've opened this PR with the minimum reproducible test case of this behavior.

### DIFF
--- a/test/clj/matcher_combinators/matchers_test.clj
+++ b/test/clj/matcher_combinators/matchers_test.clj
@@ -46,6 +46,20 @@
                                   :in-any-order)
             ::result/weight number?}))
 
+(facts "in-any-order process time grows exponentially by the number of items"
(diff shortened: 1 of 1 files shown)


Igor1201:
@philomates It seems https://github.com/nubank/matcher-combinators/pull/72 caused this issue. Could you take a look?

philomates:
addressed in https://github.com/nubank/matcher-combinators/pull/87, sorry about the delay on this

PR #81, opened 2019-08-07
introduce match-with, match-equals, match-roughly

philomates:
another attempt at https://github.com/nubank/matcher-combinators/pull/29

Inside the context of `match?` (clojure.test) / `match` (midje), data-structures are assigned default matchers, which eliminates the need to wrap data-structures with matcher-combinators when your desired matching behavior matches the defaults.

But what if your desired matching behavior deviates from the defaults?

For example, if you want to do exact map matching you need to use a log of `m/equals`:

```clojure
(deftest exact-map-matching-by-hand
  (is (match? (m/equals {:a (m/equals {:b (m/equals {:c odd?})})}))
              {:a {:b {:c 1}}})
  ;; without m/equals, the system defaults to m/embeds for maps,
  ;; which has looser matching properties
  (is (match? {:a {:b {:c odd?}}}
              {:a {:b {:c 1 :extra-c 0} :extra-b 0} :extra-a 0})))
```

This verbosity can be avoided by redefining the matcher data-type defaults

### clojure.test

You can register a custom `clojure.test` match assert expression if you are going to use it a few times:

```clojure
(defmethod clojure.test/assert-expr 'match-equals? [msg form]
  (matcher-combinators.test/build-match-assert 'match-equals? {clojure.lang.IPersistentMap m/equals} msg form))

(deftest match-equals-test
  (is (match-equals? {:a {:b {:c odd?}}}
                     {:a {:b {:c 1}}})))
```

Or if you want a one-off override of defaults, it can be done `match-with?`:

```
(deftest one-off-match-equals
  (is (match-with? {clojure.lang.IPersistentMap m/equals}
                   {:a {:b {:c odd?}}}
                   {:a {:b {:c 1}}})))
```

#### built-in matching context

- `match?`
- `match-with?`
- `match-equals?`

### midje

```clojure
(fact "match-with example"
  {:a {:b {:c odd?}}} => (match-with {clojure.lang.IPersistentMap m/equals}
                                     {:a {:b {:c odd?}}}))
(fact "match-equals example"
  {:a {:b {:c odd?}}} => (match-equals {:a {:b {:c odd?}}}))
```

#### built-in matching context

- `match`
- `match-with`
- `match-roughly`
- `match-equals`

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,6 +202,14 @@ Or if you want a one-off override of defaults, it can be done `match-with?`:
 - `match-roughly`
 - `match-equals`
 
+Or you can build your own, for example:
+
+```clojure
+(def match-equals
+  "match but using strict `equals` matching behavior for maps, even nested ones."
+  (match-with {clojure.lang.IPersistentMap matchers/equals}))
+```
+
 ## Running tests
 
 The project contains `midje`, `clojure.test`, and `cljs.test` tests.
--- a/README.md
+++ b/README.md
@@ -150,5 +150,5 @@ lein midje
 To run Clojurescript tests:
 
 ```
-lein node-test
+lein test-node
 ```
--- a/README.md
+++ b/README.md
@@ -137,6 +137,71 @@ You can extend your data-types to work with `matcher-combinators` by implemented
 
 An example of this in the wild can be seen in the `abracad` library [here](https://github.com/nubank/abracad/blob/b52e6a7114461f50bdacc2cf09a1de08f707b9f3/test/abracad/custom_types_test.clj#L15-L20).
 
+## Overriding default matchers
+
+Inside the context of `match?` (clojure.test) / `match` (midje), data-structures are assigned default matchers, which eliminates the need to wrap data-structures with matcher-combinators when your desired matching behavior matches the defaults.
+
+But what if your desired matching behavior deviates from the defaults?
+
+For example, if you want to do exact map matching you need to use a log of `m/equals`:
+
+```clojure
+(deftest exact-map-matching-by-hand
+  (is (match? (m/equals {:a (m/equals {:b (m/equals {:c odd?})})}))
+              {:a {:b {:c 1}}})
+  ;; without m/equals, the system defaults to m/embeds for maps,
+  ;; which has looser matching properties
+  (is (match? {:a {:b {:c odd?}}}
+              {:a {:b {:c 1 :extra-c 0} :extra-b 0} :extra-a 0})))
+```
+
+This verbosity can be avoided by redefining the matcher data-type defaults
+
+### clojure.test
+
+You can register a custom `clojure.test` match assert expression if you are going to use it a few times:
+
+```clojure
+(defmethod clojure.test/assert-expr 'match-equals? [msg form]
+  (matcher-combinators.test/build-match-assert 'match-equals? {clojure.lang.IPersistentMap m/equals} msg form))
+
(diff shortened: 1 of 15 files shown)


src/clj/matcher_combinators/midje.clj
```diff
@@ -86,3 +134,22 @@
          ::result/value  mismatch-val
          ::result/weight 1}))))
 
+(def match-equals
+  "match but using strict `equals` matching behavior for maps, even nested ones."
+  (match-with {clojure.lang.IPersistentMap matchers/equals}))
+
+(def match-roughly
```
philomates: still need to implement this for `clojure.test`, but it is a little more tricky because it takes in another arg

src/cljc/matcher_combinators/cljs_test.cljc
```diff
@@ -55,6 +56,13 @@
           :expected (str "The first argument of match? needs to be a matcher (implement the match protocol)")
           :actual   '~form}))))
 
+(defmethod t/assert-expr 'match-with? [_ msg form]
```
philomates: my cljs env is jacked due to rrb-vector bugs so I can't work on this right now

src/clj/matcher_combinators/midje.clj
```diff
@@ -86,3 +134,22 @@
...
+                                                    (core/roughly? expected actual delta))))]
+      (match-with
+       {java.lang.Integer    func
+        java.lang.Short      func
+        java.lang.Long       func
+        java.lang.Float      func
+        java.lang.Double     func
+        java.math.BigDecimal func
+        java.math.BigInteger func
+        clojure.lang.BigInt  func}
```
philomates: eventually I'd like to try to group similar classes so that we can do things like `{num func}` instead of the huge map :point_up:

README.md
```diff
@@ -137,6 +137,71 @@ You can extend your data-types to work with `matcher-combinators` by implemented
 
 An example of this in the wild can be seen in the `abracad` library [here](https://github.com/nubank/abracad/blob/b52e6a7114461f50bdacc2cf09a1de08f707b9f3/test/abracad/custom_types_test.clj#L15-L20).
 
+## Overriding default matchers
```
philomates: these docs aren't the greatest but I think the whole readme needs to be redone so I'm not going to refine it right now

test/clj/matcher_combinators/dispatch_test.clj
```diff
@@ -0,0 +1,68 @@
...
+  (fact "binding 1-arg match-with to new checker"
+    1 => (match-abs -1)
+    -1 => (match-abs 1)))
+
+(facts "match-equals"
+  (fact "normal loose matching passes"
+    {:a 1 :b 3 :c 1} => (match {:a 1 :b odd?}))
+  (fact "match-equals is more strict"
+    {:a 1 :b 3 :c 1} =not=> (match-equals {:a 1 :b odd?})
+    {:a 1 :b 3} => (match-equals {:a 1 :b odd?})))
```
Valbrand: Maybe a test in which one has mixed behavior would be relevant here. Let's say I want the default for a map to match with `equals` behavior, but parts of the map are huge and only some fields matter, like:

```clj
{:a {:b {:c 1}
 :d {:e 2
     :f 5
     :g 17}}
=> (match-equals {:a {:b {:c 1} 
                  :d (matchers/embeds {:e 2}})
```

PR #94, opened 2019-12-19
add an `every` matcher

dchelimsky:
The `every` matcher takes an arbitrary length sequence of matchers to be applied against an single actual value, and returns a single map with all of the matched and/or mismatched values.

The use case that motivated this was a large string with several regex matchers applied to it, so we express this

```clojure
(core/match
  (matchers/every 
    (matchers/regex #"abc")
    (matchers/regex #"def")) 
  long-string)
```

instead of this

```clojure
(core/match (matchers/regex #"abc") long-string)
(core/match (matchers/regex #"def") long-string)
```

### DIFF
--- a/src/cljc/matcher_combinators/core.cljc
+++ b/src/cljc/matcher_combinators/core.cljc
@@ -410,3 +410,21 @@
       issue
       (value-match (.toString expected)
                    (.toString actual)))))
+
+(defrecord Every [matchers]
+  Matcher
+  (match [_this actual]
+    (loop [matchers matchers
+           result   {::result/type   :match
+                     ::result/value  []
+                     ::result/weight 0}]
+      (if-let [matcher (first matchers)]
+        (let [res (match matcher actual)]
+          (recur (rest matchers)
+                 (-> result
+                     (cond->
+                         (= :mismatch (::result/type res))
+                       (assoc ::result/type :mismatch))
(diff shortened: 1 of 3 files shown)


philomates:
Nice, I like how this lets one write more concise tests!
In its current form the mismatch information for `every` deviates a little bit from other matchers in that it wraps the result in a list no matter what.

For example:
```
FAIL in (every-test) (test_test.clj:66)
expected: (match? (m/every #"123" #"233") "12345")
  actual: ["123" (mismatch #"233" "12345")]
```
is a little confusing because the `actual:` should always be the actual datastructure with mismatch annotations in any nested elements. In this case the actual is a string but the test shows a list.

I'm not sure what the best way to tackle this is, but probably adding a new record to `matcher-combaintors.model` and extending the print method to pretty print that record.

```
FAIL in (every-test) (test_test.clj:66)
expected: (match? (m/every #"123" #"233") "12345")
  actual: "123" mismatched 1 matcher: (mismatch #"233" "12345")
```

dchelimsky:
@philomates maybe this isn't as general a problem as I originally thought. Can you think of other cases in which `every` would reduce duplication and still produce useful output? I'm wondering if we'd be better served by just letting the regex matcher take multiple regexen.

dchelimsky:
This is not worth the effort IMO. Closing.

Issue #95, opened 2020-01-07
cljs-test integration hack conflicting with tools.namespace refresh


rafaeldff:
If we add matcher-combinators as source dependency of a pure clojure project (e.g. via deps `:local/root`), and try to call `clojure.tools.namespace/refresh`, it will try to load the namespace `matcher-combinators.cljs-test`, which requires `cljs.test`, leading to an error.

vemv:
In the described scenario, what are your `clojure.tools.namespace.repl/refresh-dirs`?

Generally it's good to explicitly set them. Maybe if they're not, you'll give `clojure.tools.namespace.repl/refresh` a greater workload than strictly necessary.

Hope it helps!

philomates:
I can confirm that this issue comes up with `:local/root` usage. Now with `midje` because the dependency got removed when moving to `deps.edn`.
setting `clojure.tools.namespace.repl/refresh-dirs` to the current project works to avoid this. I think this is the way to go, thanks for the suggestion @vemv

PR #96, opened 2020-01-07
Suggest ugly hack to work around tools-namespace integration problem

rafaeldff:
Suggesting an ugly hack to work around #95:

The basic idea is to try to require the namespace defining the multifn while macroexpanding, and avoid emitting the defmethod forms in case the require fails.

It has only been tested via `lein test-node`. In particular, I didn't check if test reporting was working in a clojurescript context.

### DIFF
--- a/src/cljc/matcher_combinators/cljs_test.cljc
+++ b/src/cljc/matcher_combinators/cljs_test.cljc
@@ -5,31 +5,34 @@
             [matcher-combinators.dispatch :as dispatch]
             [matcher-combinators.printer :as printer]
             [matcher-combinators.parser]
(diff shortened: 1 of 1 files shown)


dchelimsky:
This has been sitting around for over a year. Is it even an issue any longer? Can we close it?

rafaeldff:
I think that the only thing holding up merging this is testing it on a clojurescript project. Does anyone have one handy?

philomates:
as suggested in https://github.com/nubank/matcher-combinators/issues/95#issuecomment-1422898642, `clojure.tools.namespace.repl/refresh-dirs` can be used to avoid this problem, so I'm going to close this

PR #98, opened 2020-01-17
Add arglists meta to cljtest assert expressions

caioaao:
This should help out on editors. This is how it looks like in emacs:

Screenshot from 2020-01-16 19-02-05


philomates:
deployed as `1.3.1`

Issue #101, opened 2020-03-09
Unable to require only `matcher-combinators.core` and `matcher-combinators.matchers` namespaces


dchelimsky:
### Dependencies

```clojure
{:deps {nubank/matcher-combinators {:mvn/version "1.3.1"}}}
```

### Problem

I want to be able to use just the `matcher-combinators.core` and `matcher-combinators.matchers` namespaces so I can do something like this without depending on `clojure.test`:

```clojure
(ns tan.matcher-combinators
  (:require [matcher-combinators.core :refer [match]]
            [matcher-combinators.matchers :as matchers]))

(match (matchers/set-equals [uuid? uuid?])
       #{#uuid "462205d1-3cc3-4719-a980-6fff95325112"
         #uuid "55a88d99-dd0d-4fb1-8df7-4ef7cf789dd0"})
```

That code, however, produces this error:

```

  Show: Project-Only All 
  Hide: Clojure Java REPL Tooling Duplicates  (12 frames hidden)

1. Unhandled java.lang.IllegalArgumentException
   No implementation of method: :match of protocol:
   #'matcher-combinators.core/Matcher found for class:
   clojure.core$uuid_QMARK_

          core_deftype.clj:  583  clojure.core/-cache-protocol-fn
          core_deftype.clj:  575  clojure.core/-cache-protocol-fn
                 core.cljc:    8  matcher_combinators.core$eval6525$fn__6526$G__6516__6533/invoke
                 core.cljc:  258  matcher_combinators.core$matches_in_any_order_QMARK_$fn__6788/invoke
                  core.clj: 2817  clojure.core/filter/fn
              LazySeq.java:   42  clojure.lang.LazySeq/sval
              LazySeq.java:   51  clojure.lang.LazySeq/seq
              LazySeq.java:   73  clojure.lang.LazySeq/first
                   RT.java:  692  clojure.lang.RT/first
                  core.clj:   55  clojure.core/first
                  core.clj:   55  clojure.core/first
                utils.cljc:    9  matcher_combinators.utils$find_first/invokeStatic
                utils.cljc:    8  matcher_combinators.utils$find_first/invoke
                 core.cljc:  258  matcher_combinators.core$matches_in_any_order_QMARK_/invokeStatic
                 core.cljc:  249  matcher_combinators.core$matches_in_any_order_QMARK_/invoke
                 core.cljc:  281  matcher_combinators.core$matched_or_best_matchers$fn__6794/invoke
[...]

PR #102, opened 2020-03-10
add match fn to standalone ns

dchelimsky:
This is in response to #101 

Before this PR, if you want to use matcher-combinators outside of clojure.test or midje, you have to require the `matcher-combinators.parser` namespace, e.g.

```clojure
(require '[matcher-combinators.core :refer [match]]
         '[matcher-combinators.dispatch] ;; just because
)

(match 37 (count things))
```

That's fine if you happen to know about it, but it is not documented and, arguably, should not be a user-facing namespace.

With this PR you can just use the (already existing) `matcher-combinators.standalone` namespace:

```clojure
(require '[matcher-combinators.standalone :refer [match]])

(match 37 (count things))
```

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -1,4 +1,4 @@
-(defproject nubank/matcher-combinators "1.3.1"
+(defproject nubank/matcher-combinators "1.3.2"
   :description "Library for creating matcher combinator to compare nested data structures"
   :url "https://github.com/nubank/matcher-combinators"
   :license {:name "Apache License, Version 2.0"}
--- a/project.clj
+++ b/project.clj
@@ -1,4 +1,4 @@
-(defproject nubank/matcher-combinators "1.3.2"
+(defproject nubank/matcher-combinators "1.4.0"
   :description "Library for creating matcher combinator to compare nested data structures"
   :url "https://github.com/nubank/matcher-combinators"
   :license {:name "Apache License, Version 2.0"}
--- a/src/cljc/matcher_combinators/standalone.cljc
+++ b/src/cljc/matcher_combinators/standalone.cljc
@@ -3,6 +3,26 @@
             [matcher-combinators.core :as core]
             [matcher-combinators.parser]))
 
+(defn match
+  "Returns a map indicating whether the `actual` value matches the `matcher`.
+
+  `matcher` can be a matcher-combinators matcher, a predicate function of
+  actual, an expression that returns a value, or a literal value.
+
+  `actual` can be an expression that returns the actual value, or a literal.
+
+  Return map includes the following keys:
+
+  - :match/result    - either :match or :mismatch
+  - :mismatch/detail - only when :match/result is :mismatch"
+  [matcher actual]
+  (let [{:keys [matcher-combinators.result/type
+                matcher-combinators.result/value]}
+        (core/match matcher actual)]
+    (cond-> {:match/result type}
+      (= :mismatch type)
+      (assoc :mismatch/detail value))))
+
 (s/fdef match?
   :args (s/alt :partial (s/cat :matcher (partial satisfies? core/Matcher))
                :full    (s/cat :matcher (partial satisfies? core/Matcher)
@@ -10,10 +30,12 @@
   :ret boolean?)
 
 (defn match?
-  "Does the value match the provided matcher-combinator?"
+  "Given a `matcher` and `actual`, returns `true` if
+  `(match matcher actual)` results in a match. Else, returns `false.`
+
+  Given only a `matcher`, returns a function that will
+  return true or false by the same logic."
   ([matcher]
    (fn [actual] (match? matcher actual)))
   ([matcher actual]
-   (-> matcher
-       (core/match actual)
-       core/match?)))
+   (core/match? (core/match matcher actual))))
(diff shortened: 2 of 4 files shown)


src/cljc/matcher_combinators/standalone.cljc
```diff
@@ -3,14 +3,37 @@
...
+
+  `actual` can be an expression that returns the actual value, or a literal.
+
+  Return map includes the following keys:
+
+  - :matcher-combinators.result/type
+    - either :match or :mismatch
+  - :matcher-combinators.result/value
+    - when :matcher-combinators.result/type is :match, binds to the actual value,
+      else a map indicating the expected and actual values "
```
philomates: It isn't always a map, it can be either a mismatch annotation or a composite datastructure where elements inside have mismatch annotations.

for example:
 - https://github.com/nubank/matcher-combinators/blob/574945ed6c982fe6381c8ffd80f6034dd607f68a/test/clj/matcher_combinators/core_test.clj#L162
 - https://github.com/nubank/matcher-combinators/blob/574945ed6c982fe6381c8ffd80f6034dd607f68a/test/clj/matcher_combinators/core_test.clj#L62

src/cljc/matcher_combinators/standalone.cljc
```diff
@@ -3,14 +3,37 @@
...
+  [matcher actual]
+  (core/match matcher actual))
```
philomates:
```suggestion
  (dissoc (core/match matcher actual)) :matcher-combinators.result/weight)
```

src/cljc/matcher_combinators/standalone.cljc
```diff
@@ -3,14 +3,37 @@
...
   :args (s/alt :partial (s/cat :matcher (partial satisfies? core/Matcher))
                :full    (s/cat :matcher (partial satisfies? core/Matcher)
                                :actual any?))
   :ret boolean?)
 
 (defn match?
-  "Does the value match the provided matcher-combinator?"
+  "Given a `matcher` and `actual`, returns `true` if
+  :matcher-combinators.result/type in the map returned
+  by `(match matcher actual)` is :match. Else, returns `false.`
```
philomates: `:matcher-combinators.result/type` is an implementation detail and I don't know if it is helpful to talk about it here

src/cljc/matcher_combinators/standalone.cljc
philomates: I would actually prefer an implementation that doesn't expose random internal representations from `core`

So something like the following feels a little more evolvable

```clojure
(let [match-results (core/match matcher actual)]
  [(= :match match-results)
   (:matcher-combinators.result/value match-results)])
```

maybe even making it a map with proper keys, if you can think of good names

dchelimsky: When the result is `:match`, the `:value` isn't that interesting. What do you think about this:

```clojure
;; match case
{:match/result :match}

;; mismatch case
{:match/result  :mismatch
 :match/details (:matcher-combinators.result/value match-results)}
```

philomates: that sounds perfect!

dchelimsky:
@philomates suggestions handled. Pls re-review.

src/cljc/matcher_combinators/standalone.cljc
```diff
@@ -3,17 +3,42 @@
...
+
+  `actual` can be an expression that returns the actual value, or a literal.
+
+  Return map includes the following keys:
+
+  - :matcher-combinators.result/type
+    - either :match or :mismatch
+  - :matcher-combinators.result/value
+    - when :matcher-combinators.result/type is :match, binds to the actual value,
+      else a data structure with details about the mismatch."
```
philomates: this part of the doc needs to be updated

dchelimsky: Yep - noticed right after I posted that comment :-D

As I was rewriting, I realized that maybe `:mismatch/detail` would be better than `:match/detail`. WDYT? 

I also want to add tests for this before merging. Any objection to me clojure.test-ifying the standalone-test?

philomates: `:mismatch/detail` does sound better. and :+1: for tests and doing them in clojure.test :)

dchelimsky: Done.

New question: `:match/detail` currently binds to a `Mismatch` object. I'm thinking maybe that should be converted into a normal data structure. My main concern is tooling outside matcher-combinators dispatching on that type.

project.clj
```diff
@@ -1,4 +1,4 @@
-(defproject nubank/matcher-combinators "1.3.1"
+(defproject nubank/matcher-combinators "1.3.2"
```
philomates: `1.4.0` since it is a new behavior?

src/cljc/matcher_combinators/standalone.cljc
```diff
@@ -3,17 +3,39 @@
...
+
+  `matcher` can be a matcher-combinators matcher, a predicate function of
+  actual, an expression that returns a value, or a literal value.
+
+  `actual` can be an expression that returns the actual value, or a literal.
+
+  Return map includes the following keys:
+
+  - :match/result    - either :match or :mismatch
+  - :mismatch/detail - only when :match/result is :mismatch"
```
philomates:
```suggestion
  - :mismatch/detail - the actual value with mismatch annotations. Only present when :match/result is :mismatch"
```

project.clj
dchelimsky: Fine by me, but previous updates don't really follow that rule.

Issue #100, opened 2020-03-08
Provide a default implementation of Matcher protocol for Object class


alan-ghelardi:
Very often I run into situations where, due to some circumstances, the data structures that I'm comparing through matcher-combinators contain some values for which there is no implementation of the Matcher protocol and the test throws an exception (e.g. java.io.File, java.time.Instant, etc.).

Is there any reason for not providing a default implementation of the Matcher protocol for the Object class, that could match values through their equality? For example: 
```clojure
(extend-protocol matcher-combinators.core/Matcher
  Object
  (match [expected actual]
    (= expected actual)))
```
Not sure whether there are known implications against this approach.

philomates:
That is a great suggestion!
I tested it out and it didn't seem to break any tests (other than the ones that check that matching objects failed), so I went ahead and turned it into a PR: https://github.com/nubank/matcher-combinators/pull/103

alan-ghelardi:
Awesome @philomates! Thank you so much for the prompt answer.

philomates:
resolved in https://github.com/nubank/matcher-combinators/pull/103

PR #103, opened 2020-03-11
default matcher for Object

philomates:
https://github.com/nubank/matcher-combinators/issues/100

note: I don't know how to do this for clojurescript

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@ All notable changes to this project will be documented in this file. This
 change log follows the conventions of
 [keepachangelog.com](http://keepachangelog.com/).
 
-## [1.4.0]
+## [1.5.0]
 - Implement default equality matching for objects. If classes don't explicity
   implement the `match` protocol, they now default to using equality matching.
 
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,10 @@ All notable changes to this project will be documented in this file. This
 change log follows the conventions of
 [keepachangelog.com](http://keepachangelog.com/).
 
+## [1.4.0]
```
dchelimsky: 1.4.1 or 1.5.0?

src/cljc/matcher_combinators/dispatch.cljc
```diff
@@ -15,6 +15,7 @@
 ;; equals base types
 (defn nil-dispatch [expected] (matchers/equals expected))
 (defn class-dispatch [expected] (matchers/equals expected))
+(defn object-dispatch [expected] (matchers/equals expected))
 (defn integer-dispatch [expected] (matchers/equals expected))
 (defn short-dispatch [expected] (matchers/equals expected))
 (defn long-dispatch [expected] (matchers/equals expected))
```
alan-ghelardi: Didn't those more specific dispatchers become obsolete due to the Object implementation?

philomates: They are useful to leave in if you want to override the default matching behavior, such as to do rough numeric matching by default

PR #104, opened 2020-03-17
specify ~'match in macro

dchelimsky:
In the context of these macros, `(match ...` was resolving to `(matcher-combinators.parser/match ...)`, which is not a thing. It happened to work because `extend-protocol` happens to strip the namespace off of symbol, so `any-namespace-at-all/match` is the same as `match`, but we can't count on that implementation detail.

### DIFF
--- a/src/cljc/matcher_combinators/parser.cljc
+++ b/src/cljc/matcher_combinators/parser.cljc
@@ -85,14 +85,14 @@
   `(extend-protocol
     core/Matcher
      ~@(mapcat (fn [t] `(~t
-                         (match [this# actual#]
+                         (~'match [this# actual#]
                            (core/match (~matcher-builder this#) actual#)))) types)))
 
 (defmacro mimic-matcher-java-primitives [matcher-builder & type-strings]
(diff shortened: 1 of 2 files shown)


dchelimsky:
> oops, was this causing things to fail?

No, but it was working by accident. I found it because I was preparing training notes and trying to understand the internals better, so I macroexpanded a `(mimic-matcher ...)` and noticed it. It's likely that this would never ever break, but if it did we'd be really surprised by it :)

> If so, is there a nice way to write a regression test?

Not based on any failures, since there aren't any. I thought about a test with `macroexpand`, but decided not to bother. Now that you've asked about it, I'll bother (coming shortly).

philomates:
ah, cool
well I'm fine leaving out macroexpand tests; it is nice to be strict and have coverage, but I often find them very unwieldy to maintain

dchelimsky:
> well I'm fine leaving out macroexpand tests

too late ;)

dchelimsky:
I don't think this needs a release, or even a mention in the Changelog. It can just be part of the next release.

Issue #109, opened 2020-03-23
The `match?` assert-expr (used in `is`) should return `true` or `false`


dchelimsky:
``` clojure
(require '[matcher-combinators.standalone :as standalone]
         '[matcher-combinators.matchers :as matchers]
         '[matcher-combinators.clj-test]  ;; load `match?` assert-expr
         '[clojure.test :refer [is]])

(standalone/match? (matchers/equals 37) 37)
;; => true
(standalone/match? (matchers/equals 37) 42)
;; => false

(is (match? (matchers/equals 37) 37))
;; => nil
(is (match? (matchers/equals 37) 42))
;; => nil
```

This is confusing when evaluating inner expressions in order to debug.

philomates:
oh, I didn't realize that `is` expressions returned values. This definitely makes sense to fix :+1:

dchelimsky:
Fixed in #111

PR #111, opened 2020-03-24
match? in an `is` returns boolean

dchelimsky:
Solution for https://github.com/nubank/matcher-combinators/issues/109

### DIFF
--- a/test/clj/matcher_combinators/test_test.clj
+++ b/test/clj/matcher_combinators/test_test.clj
@@ -17,12 +17,22 @@
 
 (def match-data {:foo 1 :bar 2})
 
+(defmacro shhh!
+  "Evals and returns the value of body without reporting failures."
+  [& body]
+  `(with-redefs [clojure.test/do-report (constantly nil)]
+     ~@body))
+
 (deftest basic-matching
   (is (match? example-matcher example-actual)
       "In 'match?', the matcher argument comes first")
   (is (match? (m/equals example-matcher)
               (dissoc example-actual :device))
       "wrapping the matcher in 'equals' means the top level of 'actual' must have the exact same key/values")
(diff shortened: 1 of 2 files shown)


test/clj/matcher_combinators/test_test.clj
```diff
@@ -17,12 +17,22 @@
...
+
 (deftest basic-matching
   (is (match? example-matcher example-actual)
       "In 'match?', the matcher argument comes first")
   (is (match? (m/equals example-matcher)
               (dissoc example-actual :device))
       "wrapping the matcher in 'equals' means the top level of 'actual' must have the exact same key/values")
+  (is (true? (is (match? 1 1)))
+      "match? should return true for a :match")
+  (is (false? (shhh! (is (match? 1 2))))
```
markomafs: Now, that means we can assert like `(not (match? ... )) ` 

How about add a test for this? 🤔🤔🤔

dchelimsky: That is the test! Unless you want me to add a test that Clojure works as documented w/ regard to `not` `nil` and `false` ;)

dchelimsky: We _could_ add it to the docs, though.

markomafs: Works for me, mainly because we didn’t support that in the past.

dchelimsky: Actually, that won't work. `match?` is not a function - it's an `assert-expr` in clojure.test, which only works if it's the first symbol in the seq passed to `is` e.g. `(is (match? ..))`. If we were to say `(is (not (match? ...)))` we'd get an error on `match?` being unresolved.

We could add special handling for `(is (not (match? ...)))` but that would require another PR.

PR #115, opened 2020-04-16
add more clojure.test examples to README

dchelimsky:
Also add them to a namespace where we can run them


README.md
```diff
@@ -11,25 +11,91 @@ _docs:_
...
+ - `thrown-match?`: The first argument should be a throwable subclass, the second a matcher-combinators, and the third the expression being checked.
+
+For example:
+
+```clojure
+(require '[clojure.test :refer [deftest is]]
+         '[matcher-combinators.test] ;; adds support for `match?` and `thrown-match?` in `is` expressions
+         '[matcher-combinators.matchers :as m])
+
+(deftest test-matching-with-explict-matchers
```
rafaelzlisboa, line 41: `explicit`

README.md
```diff
@@ -53,7 +119,7 @@ For example:
 (fact "you can assert an exception is thrown "
   ;; Assert _some_ exception is raised and the ex-data inside satisfies the matcher
   (throw (ex-info "foo" {:foo 1 :bar 2})) => (throws-match {:foo 1})
-
+1
```
rafaelzlisboa, line 122: i think this was left here by mistake

README.md
```diff
@@ -11,25 +11,91 @@ _docs:_
...
+- Matchers for scalar and structural values
+  - Good readability supported by default interpretations of Clojure types as matchers
 - Pretty-printed diffs when the actual result doesn't match the expected matcher
-- Integrates with `clojure.test` and `midje`
-- Good readability by providing default interpretations of core Clojure data-structures as matcher combinators
+- Integration with `clojure.test` and `midje`
 
-| Midje checkers | Matcher combinators |
-| ------- | ----- |
-| ![midje checkers](doc/images/midje_check.png) | ![matcher combinators check](doc/images/matcher_check.png) |
```
rafaelzlisboa, line 26: we could ✂️ those files too

README.md
dchelimsky:
```suggestion
(deftest test-matching-with-explicit-matchers
```

README.md
dchelimsky:
```suggestion

```

README.md
```diff
@@ -11,25 +11,91 @@ _docs:_
...
+Require the `matcher-combinators.test` namespace, which will extend `clojure.test`'s `is` macro to accept the `match?` and `thrown-match?` directives.
+
+ - `match?`: The first argument should be the matcher-combinator represented the expected value, and the second argument should be the expression being checked.
+ - `thrown-match?`: The first argument should be a throwable subclass, the second a matcher-combinators, and the third the expression being checked.
+
+For example:
+
+```clojure
+(require '[clojure.test :refer [deftest is]]
+         '[matcher-combinators.test] ;; adds support for `match?` and `thrown-match?` in `is` expressions
```
FelipeCortez, line 38: oh, so this is the way it's supposed to work? I tried doing `[matcher-combinators.test :refer [match? thrown-match?]]` the other day but the `thrown-match?` refer didn't work.

FelipeCortez: actually, it does work, but only _inside_ `is`, so I think I misremembered: I was trying to refer `thrown-match?` from `matcher-combinators.standalone`, but the only def there is `match?`

FelipeCortez: still, isn't it better to explicitly mention what you're referring to play nice with linters as @philomates previously said?

README.md
```diff
@@ -11,25 +11,91 @@ _docs:_
 
 ## Motivation
 
-Clojure's built-in data structures get you a long way when trying to codify and solve difficult problems. A solid selection of core functions allow you to easily create and access core data structures. Unfortunately, this flexibility does not extend to testing: a comprehensive yet extensible way to assert that the data fits a particular structure seems to be lacking.
+Clojure's built-in data structures get you a long way when trying to codify and solve difficult problems. A solid selection of core functions allow you to easily create and access core data structures. Unfortunately, this flexibility does not extend to testing: we seem to be missinng a comprehensive yet extensible way to assert that the data fits a particular structure.
```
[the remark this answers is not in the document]
dchelimsky:
```suggestion
Clojure's built-in data structures get you a long way when trying to codify and solve difficult problems. A solid selection of core functions allow you to easily create and access core data structures. Unfortunately, this flexibility does not extend to testing: we seem to be missing a comprehensive yet extensible way to assert that the data fits a particular structure.
```

README.md
dchelimsky: @FelipeCortez if you mean include include them in a `:refer` vector, I have mixed feelings about it. `match?` is not a function, but referring it leads readers of the code to believe it is a function, as you just discovered (it works in `is` because it is a dispatch key for `clojure.test/assert-expr`).

FelipeCortez: @dchelimsky makes sense!

README.md
```diff
@@ -1,6 +1,6 @@
 # matcher-combinators
 
-Library for creating matcher combinator to compare nested data structures
+Library for making assertions about nested data structures.
```
philomates, line 3: :joy_cat: :clap:

README.md
philomates: very happy to see this horrible screenshots go away :)

README.md
```diff
@@ -11,25 +11,91 @@ _docs:_
...
+  ;; use m/set-equals to repeat predicates
+  (is (match? (m/set-equals [odd? odd? even?]) #{1 2 3})))
+
+(deftest test-matching-maps
+  ;; A map is interpreted as an `embeds` matcher, which ignores
+  ;; un-specified keys
+  (is (match? {:name/first "Alfredo"}
+              {:name/first  "Alfredo"
+               :name/last   "da Rocha Viana"
+               :name/suffix "Jr."})))
```
philomates, line 80: I mean, I'm more into Mestre Camarão these days, but works for me I guess

PR #122, opened 2020-05-21
Replace instrumentation with orchestra + add spec for partial `match?`

caioaao:
`clojure.spec.test` doesn't catch this issue since it only validates `:args`, but [orchestra](https://github.com/jeaye/orchestra) catches it.


dchelimsky:
> doesn't catch this issue

What issue?

caioaao:
@dchelimsky when you call `match?` with only one argument, it returns a function, not a boolean (as it was stated in the spec).

Issue #110, opened 2020-03-24
Add helper to discover matcher for a type


dchelimsky:
This came up in training when I was explaining that types are bound to default matchers, and it takes a bit of digging to understand which matcher is bound to which type. Something like this might help:

```clojure
(matcher-for {:this :map})
;; => (matchers/embeds {:this :map})

(matcher-for 37)
;; => (matchers/equals 37)
```

Don't know if this is feasible, but it's certainly interesting!

fernando-nubank:
The "resolution" of the matcher for a common type is made in [parser namespace](https://github.com/nubank/matcher-combinators/blob/master/src/cljc/matcher_combinators/parser.cljc).
The problem is that we only resolve while matching against a value... one idea would be adding another method to the Match protocol, but this seems like a huge change.

philomates:
one could annotate the results of a match with metadata regarding the matcher used and then grab that and print it. You would need to change matcher definitions to (optionally) include this information some how, but it could be something that is incrementally added

philomates:
cool idea btw!

fernando-nubank:
Something like 
```clojure
(matcher-used (match? {:a :map} {})) ;=> (matchers/embeds {:a :map})
```
and `matcher-used` will get this info from `meta`

philomates:
I roughly had this in my head, though who knows how it would play out

```clojure
(defn matcher-used [thing]
  (-> thing (match? nil) meta :matcher-used))
```

philomates:
addressed in https://github.com/nubank/matcher-combinators/pull/123

PR #123, opened 2020-05-21
matcher-for

dchelimsky:
Here's a proposed solution for #110 that @fernando-nubank and I cooked up.

`matcher-for` creates a matcher for you, so you can inspect it:

```clojure
(matchers/matcher-for 37)
;; => matcher_combinators.core.Value#{:expected 37}
```

And then you can also then use it!

```clojure
(let [m (matchers/matcher-for 37)]
  (standalone/match? m 37))
;; => true
```

It even works within `match-with?`:

```clojure
(deftest matcher-for-works-within-match-with
  (is (match-with? {java.lang.Long greater-than-matcher}
                   (matchers/matcher-for 4)
                   5)))
```

This PR also moves some tests around and converts more from midje to clojure.test (and adds a generative for `matcher-for`.

## TODO

- [x] get feedback
- [x] update README
- [x] ensure works with cljs
- [x] change `Matcher/match` to `Matcher/-match` (internal)
  - [x] and add `core/match` to make it a non-breaking change for standard use
- [x] update CHANGELOG
  - note that it is a breaking change for custom implementations of `Matcher`
- [x] bump major version

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,14 +135,37 @@ Note that you can also use the `match` checker to match arguments within midje's
 
 ## Matchers
 
-### default matchers
+### Default matchers
 
-If a data-structure isn't wrapped in a specific matcher-combinator the default interpretation is:
-- map: `embeds`
+When an expected value isn't wrapped in a specific matcher the default interpretation is:
+- number, date, and other scalar values: `equals`
+- regex: `regex`
 - sequential: `equals`
 - set: `equals`
-- number, date, and other base data-structure: `equals`
-- regex: `regex`
+- map: `embeds`
+
+You can use the `matcher-for` function to discover which matcher would be used
+for a specific value, e.g.
+
+``` clojure
+(require '[matcher-combinators.matchers :as matchers])
+
+(matcher-combinators/matcher-for {:this :map})
+;; => {:expected {:this :map}}
+
+(class *1)
+;; => matcher_combinators.core.EmbedsMap
+```
+
+You can then use that matcher in place of a value, e.g.
+
+``` clojure
+(require '[matcher-combinators.standalone :as matcher-combinators])
+(matcher-combinators/match
+  (matchers/matcher-for {:this :map})
+  {:this :map :and :then-some})
+;; => {:match/result :match}
+```
 
 ### built-in matchers
 
--- a/README.md
+++ b/README.md
@@ -150,7 +150,7 @@ for a specific value, e.g.
 ``` clojure
 (require '[matcher-combinators.matchers :as matchers])
 
-(matcher-combinators/matcher-for {:this :map})
+(matchers/matcher-for {:this :map})
 ;; => {:expected {:this :map}}
 
 (class *1)
--- a/src/cljc/matcher_combinators/core.cljc
+++ b/src/cljc/matcher_combinators/core.cljc
@@ -6,16 +6,30 @@
             [matcher-combinators.utils :as utils]))
 
 (defprotocol Matcher
-  "For matching expected and actual values, providing helpful mismatch info on
-  unsucessful matches"
+  "For internal use. Type-specific implementations for finding matchers for
+  expected values and matching them against expected values."
   (-matcher-for [expected]
     "Do not call directly. Implementation for matcher-combinators.matchers/matcher-for.")
-  (match [this actual]
-    "determine if a concrete `actual` value satisfies this matcher"))
+  (-match [this actual]
+    "Do not call directly. Implementation for matcher-combinators.core/match."))
+
+(defn match
+  "For internal use. Returns a map indicating whether the `actual` value matches `expected`.
+
+  `expected` can be the expected value, a matcher, or a predicate fn of actual.
+
+  Return map includes the following keys:
+
+  - :matcher-combinators.result/type  - either :match or :mismatch
+  - :matcher-combinators.result/value - the actual value with mismatch annotations.
+                                        Only present when :match/result is :mismatch"
+  [expected actual]
+  (-match expected actual))
 
 (s/fdef match?
   :args (s/cat :match-result ::result/result)
   :ret boolean?)
+
 (defn match? [{::result/keys [type]}]
   (= :match type))
 
@@ -40,7 +54,7 @@
 (defrecord Value [expected]
   Matcher
   (-matcher-for [this] this)
-  (match [_ actual]
+  (-match [_ actual]
     (value-match expected actual)))
 
 (defn- validate-input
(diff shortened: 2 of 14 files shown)


markomafs:
Love it. 
It will help us to identify whenever we have some corner cases when matches fail. 
👏

dchelimsky:
@philomates @rafaeldff we'd like to get feedback from one or both of you before we merge this.

src/cljc/matcher_combinators/dispatch.cljc
```diff
@@ -1,4 +1,6 @@
 (ns matcher-combinators.dispatch
+  "core/Matchers$match gets bound to dispatch functions defined in this namespace
+  so that we can override them at runtime."
```
philomates, line 3: :+1: 

> This namespace provides a level of indirection between the matcher assigned to a datatype and the matcher implementation. This indirection allows for redefinition at runtime, necessary for the `match-with` feature

Is how I would write it, not sure if any of that seems useful to incorporate

src/cljc/matcher_combinators/parser.cljc
```diff
@@ -140,4 +145,5 @@
 (mimic-matcher dispatch/repeat-dispatch Repeat)
 (mimic-matcher dispatch/lazy-seq-dispatch LazySeq)
 (mimic-matcher dispatch/array-seq-dispatch ArraySeq)
-(mimic-matcher dispatch/pattern-dispatch Pattern)))
+(mimic-matcher dispatch/pattern-dispatch Pattern)
+))
```
philomates, line 149: :nail_care:

src/cljc/matcher_combinators/core.cljc
```diff
@@ -8,6 +8,7 @@
 (defprotocol Matcher
   "For matching expected and actual values, providing helpful mismatch info on
   unsucessful matches"
+  (matcher-for [this])
```
philomates, line 11: might be nice to add a docstring

test/clj/matcher_combinators/core_test.clj
```diff
@@ -572,4 +579,36 @@
                                   :in-any-order)
             ::result/weight 2}))
 
+(deftest matcher-for-special-cases
+  (testing "matcher for a fn is a fn"
+    (is (fn? (core/matcher-for (fn [])))))
```
philomates, line 584: I feel like this is a little misleading. As far as I recall, the fact that it is a function is just because of how we do dispatch for function matching.
For instance, it could (and maybe should) be reimplemented like this:

```
diff --git a/src/cljc/matcher_combinators/dispatch.cljc b/src/cljc/matcher_combinators/dispatch.cljc
index c4f3d65..4c475ba 100644
--- a/src/cljc/matcher_combinators/dispatch.cljc
+++ b/src/cljc/matcher_combinators/dispatch.cljc
@@ -58,7 +58,8 @@
 
 ;; other
 (defn pattern-dispatch [expected] (matchers/regex expected))
-(defn function-dispatch [expected] (partial core/match-pred expected (str "predicate: " expected)))
+(defn function-dispatch [expected] (core/->PredMatcher expected
+                                                       (str "predicate: " expected)))
 
 (def type->dispatch
   #?(:cljs {}
diff --git a/src/cljc/matcher_combinators/parser.cljc b/src/cljc/matcher_combinators/parser.cljc
index 58c5454..5c54aeb 100644
--- a/src/cljc/matcher_combinators/parser.cljc
+++ b/src/cljc/matcher_combinators/parser.cljc
@@ -18,7 +18,7 @@
   ;; function as predicate
   function
   (match [this actual]
-    ((dispatch/function-dispatch this) actual))
+    (core/match (dispatch/function-dispatch this) actual))
 
   ;; equals base types
   nil
@@ -107,7 +107,7 @@
   core/Matcher
   (matcher-for [this] (dispatch/function-dispatch this))
   (match [this actual]
-    ((dispatch/function-dispatch this) actual)))
+    (core/match (dispatch/function-dispatch this) actual)))
 
 (mimic-matcher dispatch/nil-dispatch nil)
 (mimic-matcher dispatch/class-dispatch java.lang.Class)
```

test/clj/matcher_combinators/matchers_test.clj
```diff
@@ -38,177 +37,177 @@
...
-         (map vals))
-    => (has every? one-mismatch?))
-
-  (fact "in-any-order minimization doesn't find the match ordering that leads
+(deftest in-any-order
+  (testing "matcher ordering with maximum matchings for diff"
+    (is (match?
+         {::result/type   :mismatch
+          ::result/value  (m/embeds [a-nested-map {:id map? :model mismatch?}])
+          ::result/weight number?}
```
philomates, line 48: if you want a behaviorally equivalent test, you need to wrap this map in an `equals` matcher

test/clj/matcher_combinators/matchers_test.clj
```diff
@@ -38,177 +37,177 @@
...
-         ::result/value
-         (map vals))
-    => (has every? one-mismatch?))
-
-  (fact "in-any-order minimization doesn't find the match ordering that leads
+(deftest in-any-order
+  (testing "matcher ordering with maximum matchings for diff"
+    (is (match?
+         {::result/type   :mismatch
+          ::result/value  (m/embeds [a-nested-map {:id map? :model mismatch?}])
```
philomates, line 47: I think it should be `in-any-order` matcher, not `embeds`

src/cljc/matcher_combinators/dispatch.cljc
dchelimsky: I combined ideas from this with what was  there before.

test/clj/matcher_combinators/core_test.clj
dchelimsky: I merged this patch and also modified it further, in a separate commit, to go through a matcher fn to align with the other dispatch fns.

dchelimsky:
Thanks for the review @philomates. I've addressed all your comments. Can I get a follow up review and maybe an approval?

dchelimsky:
> Before merging can you add examples to the README?

It's in the TODO list in the description ;)

README.md
```diff
@@ -135,14 +135,37 @@ Note that you can also use the `match` checker to match arguments within midje's
...
-- regex: `regex`
+- map: `embeds`
+
+You can use the `matcher-for` function to discover which matcher would be used
+for a specific value, e.g.
+
+``` clojure
+(require '[matcher-combinators.matchers :as matchers])
+
+(matcher-combinators/matcher-for {:this :map})
```
dchelimsky, line 153:
```suggestion
(matchers/matcher-for {:this :map})
```

Issue #116, opened 2020-04-21
Matching against `java.util.HashSet` doesn't work


gabrielgiussi:
### The problem

I spend some time debugging why a test passed if I were using `=` but failed if I were using `match?` instead, what doesn't make sense since `match?` should be more permissive.
The output was something like this, which was confusing for me:
```clojure
#:matcher-combinators.result{:type :mismatch,
                             :value #matcher_combinators.model.Mismatch{:expected #{1 2}, :actual #{1 2}},
                             :weight 1}
```

After some debugging I found the cause and it was that the expected set was a `java.util.HashSet` and this doesn't pass the predicate `set?` used in the `validate-input`
https://github.com/nubank/matcher-combinators/blob/93eabde02940d8bf7a2e15fc0f1e77b7aefe97c0/src/cljc/matcher_combinators/core.cljc#L333-L337

If you use the java set as the actual result it will fail but it is clear what is happening, although I don't know if the user should be faced with this kind of error, there is a better feedback that we could provide to they?
```clojure
No implementation of method: :match of protocol: #'matcher-combinators.core/Matcher found for class: java.util.HashSet
```

### What to do?

I'm not sure if `match?` should handle this case, because it requires too much work, but at least it would be nice to provide a better error message.

### Namespace to reproduce this stuff

```clojure
(ns spaceteam.db.datomic.a
  (:require [matcher-combinators.standalone :refer [match?]]
            [matcher-combinators.core :as core]))


(def clojure-set #{1 2})

(def java-set (doto (new java.util.HashSet) (.add 1) (.add 2)))

(= clojure-set java-set)
; => true

(match? clojure-set java-set)
; => false

#_(match? java-set clojure-set)
; This fails with 'No implementation of method: :match of protocol: #'matcher-combinators.core/Matcher found for class: java.util.HashSet'

(core/match clojure-set java-set)
; => #:matcher-combinators.result{:type :mismatch,
;                             :value #matcher_combinators.model.Mismatch{:expected #{1 2}, :actual #{1 2}},
;                             :weight 1} 
```

dchelimsky:
> I'm not sure if match? should handle this case, because it requires too much work

`matcher-combinators` already has explicit support for some java types here: https://github.com/nubank/matcher-combinators/blob/93eabde02940d8bf7a2e15fc0f1e77b7aefe97c0/src/cljc/matcher_combinators/dispatch.cljc#L63-L97

[...]

PR #125, opened 2020-05-29
Use set matcher if the expected is a clojure set or a java set

gabrielgiussi:
Solution for issue https://github.com/nubank/matcher-combinators/issues/116.

Clojure only recognizes `IPersistentSet` as `set?` because `java.util.Set` has no immutability semantic baked in. For the purposes of matching, we only need to read the sets, so this PR applies set matching semantics to instances of `java.util.Set` as well as `clojure.lang.IPersistentSet`.

### DIFF
--- a/src/cljc/matcher_combinators/core.cljc
+++ b/src/cljc/matcher_combinators/core.cljc
@@ -346,6 +346,12 @@
       issue
       (match-any-order expected actual false))))
 
+(defn- matchable-set?
+  "Clojure's set functions expect clojure.lang.IPersistentSet, but
+  matching works just fine with java.util.Set as well."
+  [s]
+  (or (set? s) (instance? java.util.Set s)))
+
 (defrecord SetEquals [expected accept-seq?]
   Matcher
   (-matcher-for [this] this)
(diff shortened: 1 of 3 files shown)


fernando-nubank:
> Do you think it is a good solution for this issue to modify the predicate so it uses the set matcher if the expected value is a java set?

Makes sense to me - although I don't see anyone using a Java Set as expected value

src/cljc/matcher_combinators/core.cljc
```diff
@@ -359,7 +359,7 @@
                                      "set or sequential")
                      (validate-input expected
                                      actual
-                                     set?
+                                     (fn [s] (or (set? s) (instance? java.util.Set s)))
```
dchelimsky, line 362: I'm not sure about this solution yet overall, but if we're going to do it, we should include this in the `accept-seq?` case above as well.

dchelimsky:
> Makes sense to me - although I don't see anyone using a Java Set as expected value

Perhaps not when using a literal, but what if you're deriving the actual value from another function or a generator?

src/cljc/matcher_combinators/core.cljc
philomates: Maybe we can look to see what seq/set functions in clojure do when given java sets/seqs? If they interoperate, then we could do it here to. If they don't, then maybe there is a specific reason for that.

dchelimsky: > If they interoperate, then we could do it here to. If they don't, then maybe there is a specific reason for that.

Anecdotally, it looks like set functions like `clojure.set/difference` throw errors on HashSets, e.g.

(require '[clojure.set :as set])

(set/difference #{:a :b} #{:b :c})
;; => #{:a}

(set/difference
 (doto (java.util.HashSet.) (.add :a) (.add :b))
 (doto (java.util.HashSet.) (.add :b) (.add :c)))
;; => 1. Unhandled java.lang.ClassCastException
;;       class java.util.HashSet cannot be cast to class
;;       clojure.lang.IPersistentSet
```

That said, I'd guess that the context for that choice includes considerations that we don't really need to make (general performance and semantic concerns), whereas this is a specific use case.

Comment (dchelimsky) [reply src/cljc/matcher_combinators/core.cljc:362]: Also:

```clojure
(set/difference
 #{:a :b}
 (doto (java.util.HashSet.) (.add :b) (.add :c)))
;; => #{:a}

(set/difference
 (doto (java.util.HashSet.) (.add :a) (.add :b))
 #{:b :c})
;; => 1. Unhandled java.lang.ClassCastException
```

dchelimsky: @philomates I think the reason that only instances of `IPersistentSet` return true for `set?` is that `java.util.Set` instances are mutable. For the purposes of matching, since we only ever read them, I think it's fine to treat `java.util.Set` as matchable. WDYT? I've updated the branch to cover the other set-matching cases.

PR #127, opened 2020-06-03
redesign match fns

dchelimsky:

They all have different semantics, which is cause for much confusion.

This PR addresses this as follows:

* `match?` is now the arity-2 version
  * the arity-1 version is present but deprecated
* new `indicates-match?` fn replaces the arity-1 version of `match?


rafaelzlisboa:
I like it! 👍

dchelimsky:
Converting back to a Draft. I just discovered that there is a lot of usage of the arity-1 version of `standalone/match?` within Nubank. Need to re-think.

dchelimsky:
We went with #128 instead.

PR #128, opened 2020-06-03
Add `indicates-match?` (and deprecate `core/match?`)

dchelimsky:

`core/match?` receives a match result (map) while the others receive expected/actual values.

This PR eliminates the confusion by deprecating `core/match?` and replacing it with a new `indicates-match?` function.

This is not a breaking change, and will have almost no fan-out within Nubank.

### DIFF
--- a/test/clj/matcher_combinators/core_test.clj
+++ b/test/clj/matcher_combinators/core_test.clj
@@ -14,23 +14,10 @@
             [matcher-combinators.result :as result]
             [matcher-combinators.dispatch :as dispatch]
             [matcher-combinators.parser]
-            [matcher-combinators.standalone :as standalone]))
+            [matcher-combinators.standalone :as standalone]
+            [matcher-combinators.test-helpers :as test-helpers :refer [gen-any-equatable]]))
 
-(spec.test/instrument)
-
-(use-fixtures :once
-  (fn [t]
-    (spec.test/instrument)
-    (t)
-    (spec.test/unstrument)))
-
-(def gen-any-equatable
-  (gen/such-that
-   (fn [v]
-     (every? (fn [node] (or (not (set? node))
-                            (not (contains? node false))))
-             (tree-seq coll? #(if (map? %) (keys %) %) v)))
-   gen/any-equatable))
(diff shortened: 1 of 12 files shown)


fernando-nubank:
I like this approach but `indicates-match` feels like a strange name =/

dchelimsky:
> I like this approach but `indicates-match` feels like a strange name =/

What do you suggest?

fernando-nubank:
`match?` receives `actual` and `expected` and `indicates-match?` receives a `match-result`, so
```clojure
; currently 
(indicates-match? (match? odd? 1)) ;=> true

;proposal
(matched? (match? odd? 1)) ;=> true
```
because `match-result` is the result of an already executed `match?`

test/clj/matcher_combinators/core_test.clj
```diff
@@ -96,6 +83,8 @@
                     ::result/value (model/->Mismatch expected actual)}
                    res))))
 
+(spec.test/instrument)
```
rafaelzlisboa, line 86: ✂️

dchelimsky: It was there already at the top of the namespace, to ensure that the midje facts are run with instrumentation on. I think it should stay. WDYT now that you have the context?

rafaelzlisboa:
@fernando-nubank I like the fact that `indicates-match?` reads _very_ different from `match?`. I think we should move away from having different fns with very similar names (`match`, `match?` etc)

dchelimsky:
I, personally, prefer `indicates-match?` over `matched?` for this reason:

`matched?` makes you think of an event that happened in the past: did it match back then when it happened?

`indicates-match?` says exactly what it means: does the thing you gave me indicate a match? It doesn't care how or when the thing got made.

fernando-nubank:
@rafaelzlisboa I would like to have a single worded function, but you got a point. Let's keep it then

dchelimsky:
> I would like to have a single worded function,

Keep in mind that this is really for internal use, or in support of building custom matcher functions in other libs. It's not something people will type very much, if ever, but when they see it, they're very unlikely to confuse it with `match?` ;)

PR #131, opened 2020-06-04
add Matcher impl for Symbol in cljs

dchelimsky:
This PR adds `Symbol` to the cljs types that get `Matcher` implementations.

Also, converts some cljs tests to property based tests (which was actually discovered that `Symbol` was missing).


dchelimsky:
NOTE: ignore the ci/circleci failure. We're running CI via a github action now (Clojure CI / build) but need somebody with proper permissions to disable circleci (@philomates ?).

Issue #124, opened 2020-05-27
(match #{false} #{false}) fails


dchelimsky:
```clojure
(core/match #{false} #{false})
;; => {:matcher-combinators.result/type :mismatch,
;;     :matcher-combinators.result/value #{false {:actual false}},
;;     :matcher-combinators.result/weight 1}
```

Also, because `in-any-order` uses the same underlying mechanism as set matchers:

```clojure
(core/match (m/in-any-order [false]) [false])
;; => {:matcher-combinators.result/type :mismatch,
;;     :matcher-combinators.result/value (false {:actual false}),
;;     :matcher-combinators.result/weight 1}
```

dchelimsky:
FYI - I discovered this adding a property test for the `equals` matcher:

```clojure
(defspec equals-matchers
  {:num-tests 100
   :max-size  10}
  (prop/for-all [v gen/any-equatable]
                (core/match?
                 (core/match (matchers/equals v) v))))

;; produces

{:shrunk
 {:total-nodes-visited 11,
  :depth 1,
  :pass? false,
  :result false,
  :result-data nil,
  :time-shrinking-ms 1,
  :smallest [#{false}]},
 :failed-after-ms 13,
 :num-tests 36,
 :seed 1590595380500,
 :fail [#{false \:}],
 :result false,
 :result-data nil,
 :failing-size 5,
 :pass? false,
 :test-var "equals-matchers"}
```

philomates:
ha, this is an awesome bug! I'm having fun imagining what the buggy code looks like :)
yay generative tests, was actually wondering how you came across this

philomates:
addressed in https://github.com/nubank/matcher-combinators/pull/133

PR #133, opened 2020-06-16
fix issue matching `false` value in certain contexts

philomates:
address https://github.com/nubank/matcher-combinators/issues/124


dchelimsky:
@philomates I pushed a commit that removes a test hack we had in place to ignore this bug: https://github.com/nubank/matcher-combinators/commit/e9dbf53dfbc2e38dec64a13dd4128d8cfe55e5ee. Can I just push it to this branch?

Issue #129, opened 2020-06-03
match-with? as matcher, instead of assert expression


lennartbuit:
Hey, loving this library! One thing I just stumbled upon was wanting to `match?` with a nested structure, but asserting that it was exactly equals.

Now, this can be done with `match-with?` (clojure test user here). However, we usually write tests as long `are` blocks with many different cases exercising the same function. Sometimes that means that you need to special case a matcher, e.g. in the general case I need `m/embeds`, but in this particular case I need `m/equals`. With this particular combination, its a bit unfortunate that `match-with?` is an assert expression, as that means breaking out of the current `are`, writing a new `are` with `match-with?`, etc.

So I was thinking, would it be possible to expose `match-with?` not just as an assert expression, but also as a matcher (like `m/equals`), and if so, is that desirable :)?

dchelimsky:
The nice thing about assert expressions is you just load a namespace and there they are. The bad thing about assert expressions is you just load a namespace and there they are.

I could see leaving the assert expressions in place, and adding `match-with?`, `match-equals?` and `match-roughly?` to the `standalone` namespace, but it would be on you to namespace them in an `is` expression. I need to get some feedback from other maintainers before pursuing this, and make no guarantees we'll do it at this point.

In the mean time, one alternative to breaking out of `are` _might_ (I may be missing context) be to add the matcher you want to use to the `are` arguments and then wrap it around the `are` predicate, e.g. 

```clojure
(are [matcher expected actual]
  (match? (matcher expected) actual)
  m/embeds expected1 actual
  m/equals expected2 actual
  ,,,)
```

Would that work for now?

philomates:
Hey @lennartbuit, cool to hear you are enjoying the lib!
I agree that having the only way to set the default matchers at the outer `match?` level is a bit constraining. Technically speaking, it should be possible to introduce constructs that would recursively alter the default matchers for lexically scoped substructures. I'm not exactly sure why I didn't do that when I implemented `match-with?` but thinking about it now it feels hard to "get right" and is definitely a trade-off of power/expressiveness and readability/usability.

That said, if we can think of compelling use cases and a cohesive design, it would be cool to try!

Brainstorming what this might look like:
 - remove `match-with?`, `match-equals?`, etc and just leave `match?`
 - add "context modifiers" (horrible name), which are sorta like matchers, but apply recursively

```clojure
;; so
(match-equals?
  [{:a [{:b [1]}]}]
  whatever)

;; would become
(match?
[...]

PR #134, opened 2020-06-25
match-with matcher

dchelimsky:
This PR, which addresses issue #129 introduces a `match-with` matcher, which has a similar API to the `match-with?` assert expression, but works directly on `expected` values. You can use it anywhere in an arbitrarily deeply nested structure, and overrides can be, themselves, overridden in deeper scopes with nested `match-with` matchers, e.g.

```clojure
(deftest example
  (is (match? {:a (m/match-with
                   {clojure.lang.IPersistentMap m/equals}
                   {:b
                    (m/match-with
                     {clojure.lang.IPersistentMap m/embeds
                      clojure.lang.IPersistentVector m/embeds}
                     {:c [odd? even?]})})}
              {:a {:b {:c [1 2 3]
                       :d :e}}
               :f :g})))
```

This also implements the midje match-with checker and the clojure.test match-with? assert-expr in terms of the new match-with matcher, allowing us to bind types directly to their matchers without the dispatch mechanism and eliminate `with-redefs`.


dchelimsky:
One thing that bugs me about this addition is that it introduces more `with-redefs`. Just a note that I'm exploring an alternate implementation of dispatch that would eliminate the need for `with-redefs`. I'll follow up if anything useful comes of it.

markomafs:
> One thing that bugs me about this addition is that it introduces more `with-redefs`. Just a note that I'm exploring an alternate implementation of dispatch that would eliminate the need for `with-redefs`. I'll follow up if anything useful comes of it.

Do you think i would work with meta data? 
Checking for the meta data to define the “dispatchers” . 

Another option would be dynamic var. 

Wdyt?

dchelimsky:
> > One thing that bugs me about this addition is that it introduces more `with-redefs`. Just a note that I'm exploring an alternate implementation of dispatch that would eliminate the need for `with-redefs`. I'll follow up if anything useful comes of it.
> 
> Do you think i would work with meta data?
> Checking for the meta data to define the “dispatchers” .
> 
> Another option would be dynamic var.
> 
> Wdyt?

Metadata of which Var?

I know of at least one person interested in this repo who will object to dynamic vars :)

I'm working on solving this with an additional function on the `Matcher` protocol that you can hand the type->matcher map directly. No redefs. No metadata. No dynamic vars. Should have something early next week.

dchelimsky:
@markomafs I reimplemented this without any redefs, and was able to eliminate the dispatch layer in the process.

/cc @nubank/eng-prod-test @philomates @sovelten

src/cljc/matcher_combinators/parser.cljc
```diff
@@ -18,136 +17,133 @@
...
-    (core/match (dispatch/function-dispatch this) actual)))
+  (-matcher-for
+    ([this] (matchers/pred this))
+    ([this t->m] (matchers/pred this)))
+  (-match [this actual]
+    (core/match (matchers/pred this) actual)))
+
+(defmacro mimic-matcher [matcher t]
+  `(extend-type ~(if (string? t)
+                   (symbol t)  ;; java primitives
```
fernando-nubank, line 97: Are we using this?

src/cljc/matcher_combinators/matchers.cljc
```diff
@@ -71,9 +71,79 @@
...
+         value
+
+         ;; TODO: all of the built in matchers are records, but users
+         ;; define matchers by reifying the Matcher protocol, so this
+         ;; would break down. Also, what if a user's domain includes a
+         ;; record with an `:expected` key? Ideally, we should have
+         ;; some other marker to identify a matcher object, and document
+         ;; it in terms of "your custom Matcher implementations must do
+         ;; x in order to particpate in match-with"
+         (and (record? value) (map? (:expected value)))
```
fernando-nubank, line 121: Where is `:expected` being assigned?

dchelimsky: It's part of all of the matcher record implementations.

fernando-nubank: Oh, it is the parameter of all records. Unfortunately we cannot use a namespaced keyword :(
https://clojure.atlassian.net/browse/CLJ-1938

src/cljc/matcher_combinators/matchers.cljc
```diff
@@ -71,9 +71,102 @@
...
+
+         (map? value)
+         (matcher-for (match-with-values value overrides) overrides)
+
+         (coll? value)
+         (matcher-for (match-with-elements value overrides) overrides)
+
+         :else
+         (matcher-for value overrides))
+   assoc :match-with? true))
```
philomates, line 172: do you think it is worth namespacing this keyword? It always feels nice to do this type of thing for global ops like messing with meta

src/cljc/matcher_combinators/matchers.cljc
```diff
@@ -71,9 +71,102 @@
...
+          (empty coll)
+          coll))
+
+(defn match-with [overrides value]
+  (vary-meta
+   (cond (:match-with? (meta value))
+         value
+
+         (fn? value)
+         value
```
philomates, line 149: folks might want to do exact function matching using `match-with`

like the following should pass but it doesn't currently
```
(deftest fn-exact
  (is (match? (m/equals odd?)
              odd?))
  (is (match?
       (m/match-with [fn? m/equals] odd?)
       odd?)))
```

src/cljc/matcher_combinators/matchers.cljc
```diff
@@ -71,9 +71,102 @@
...
+
+         ;; TODO: all of the built in matchers are records, but users
+         ;; define matchers by reifying the Matcher protocol, so this
+         ;; would break down. Also, what if a user's domain includes a
+         ;; record with an `:expected` key? Ideally, we should have
+         ;; some other marker to identify a matcher object, and document
+         ;; it in terms of "your custom Matcher implementations must do
+         ;; x in order to particpate in match-with"
```
philomates, line 157: we could define an interface for accessing the expected. Then you wouldn't be bound by the record arg name, in a sense. But it is crufty :)

src/cljc/matcher_combinators/matchers.cljc
```diff
@@ -71,9 +71,102 @@
...
+
+(defn- match-with-elements [coll overrides]
+  (reduce (fn [c v]
+            (conj c (match-with overrides v)))
+          (empty coll)
+          coll))
+
+(defn match-with [overrides value]
+  (vary-meta
+   (cond (:match-with? (meta value))
```
philomates, line 145: I think a comment here to say that it is to skip overriding nested `match-with` calls would be useful

src/cljc/matcher_combinators/matchers.cljc
```diff
@@ -71,9 +71,102 @@
...
+         ;; define matchers by reifying the Matcher protocol, so this
+         ;; would break down. Also, what if a user's domain includes a
+         ;; record with an `:expected` key? Ideally, we should have
+         ;; some other marker to identify a matcher object, and document
+         ;; it in terms of "your custom Matcher implementations must do
+         ;; x in order to particpate in match-with"
+         (and (record? value) (map? (:expected value)))
+         (update value :expected match-with-values overrides)
+
+         (and (record? value) (coll? (:expected value)))
```
philomates, line 161: very cool trick to distinguish between clojure datastructures like maps and custom matchers

src/cljc/matcher_combinators/matchers.cljc
```diff
@@ -71,9 +71,102 @@
...
+             {}
+             m))
+
+(defn- match-with-elements [coll overrides]
+  (reduce (fn [c v]
+            (conj c (match-with overrides v)))
+          (empty coll)
+          coll))
+
+(defn match-with [overrides value]
```
philomates, line 143: a docstring would be good here

src/cljc/matcher_combinators/matchers.cljc
dchelimsky: That's because we treat functions special to support predicate matchers. I don't know how we can do both.

src/cljc/matcher_combinators/matchers.cljc
```diff
@@ -71,9 +71,102 @@
...
+  "Internal use only. Iterates through pred->matcher-overrides and
+  returns the value (a matcher) bound to the first pred that returns
+  true for value. If no override is found, returns the default matcher
+  for value.
+
+  The legacy API called for a map of type->matcher, which is still
+  supported by wrapping types in (instance? type %) predicates."
+  [value pred->matcher-overrides]
+  (or (->> (format-overrides pred->matcher-overrides)
+           (filter (fn [[class-or-pred matcher]] (when ((->pred class-or-pred) value) matcher)))
```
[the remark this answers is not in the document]
dchelimsky: It only happens once now.

src/cljc/matcher_combinators/parser.cljc
```diff
@@ -18,136 +12,102 @@
...
+      ([this#]
+       (~matcher this#))
+      ([this# t->m#]
+       (let [m# (matchers/lookup-matcher this# t->m#)]
+         (m# this#))))
+     (~'-match [this# actual#]
+      (core/match (~matcher this#) actual#))))
+
+;; default for most objects
+(mimic-matcher matchers/equals Object)
```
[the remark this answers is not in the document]
dchelimsky: 🎉

src/cljc/matcher_combinators/matchers.cljc
dchelimsky: thx

src/cljc/matcher_combinators/matchers.cljc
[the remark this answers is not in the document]
dchelimsky: FYI - I see `update-keys` (you called it `map-keys`) and `update-vals` in a lot of projects. The first time I saw it was https://github.com/jaycfields/jry/blob/d79cc8ec552c11122001bc1dd01b9ef6c251a9fb/src/clojure/jry.clj#L67-L71

fernando-nubank:
Is this test supposed to pass?

```clojure
(is (match?
        (m/match-with [map?    m/equals
                       vector? m/embeds]
          {:a (m/match-with [map? m/embeds]
                {:vec [1 2]
                 :a/X :a/X1})})
        {:a {:a/X :a/X1
             :a/Y :a/Y1
             :vec [1 2 3]}}))
```

this declares a matcher for `vector?` that is not used in the inner matcher.

dchelimsky:
> Is this test supposed to pass?
> 
> ```clojure
> (is (match?
>         (m/match-with [map?    m/equals
>                        vector? m/embeds]
>           {:a (m/match-with [map? m/embeds]
>                 {:vec [1 2]
>                  :a/X :a/X1})})
>         {:a {:a/X :a/X1
>              :a/Y :a/Y1
>              :vec [1 2 3]}}))
> ```
> 
> this declares a matcher for `vector?` that is not used in the inner matcher.

@fernando-nubank No. The implementation doesn't support merging of match-with scopes. I documented this in the docstring for match-with.

PR #135, opened 2020-06-30
Add within-delta matcher

dchelimsky:
This PR adds a new `within-delta` matcher, intended to replace the midje `match-roughly` checker and clojure.test `match-roughly?` assert expr

```clojure
;; before
(is (match-roughly? <delta> <expected> <actual>))

;; after
(is (match? (within-delta <delta> <expected> <actual>))
;; or (for nested values)
(is (match? (match-with [number? (within-delta <delta>)] <expected>)
             <actual>))
```

### DIFF
--- a/src/cljc/matcher_combinators/utils.cljc
+++ b/src/cljc/matcher_combinators/utils.cljc
@@ -1,9 +1,17 @@
-(ns matcher-combinators.utils)
+(ns matcher-combinators.utils
+  "Internal use only. Subject (and likely) to change.")
 
-(defn roughly? [expected actual delta]
-  (and (number? actual)
-       (>= expected (- actual delta))
-       (<= expected (+ actual delta))))
+(defn processable-number? [v]
+  (and (number? v)
+       (try
+         (and (not (Double/isInfinite v))
+              (not (Double/isNaN v)))
+         (catch Exception _ false))))
+
+(defn within-delta? [delta expected actual]
+  (and (processable-number? actual)
+       (>= expected (- actual (Math/abs delta)))
+       (<= expected (+ actual (Math/abs delta)))))
 
 (defn find-first [pred coll]
   (->> coll (filter pred) first))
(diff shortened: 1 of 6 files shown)


dchelimsky:
This is implemented in terms of `match-with`, so applies to nested contexts. I'm wondering if it would be clearer to make it only apply to it's direct value. This would make the `match-roughly?` replacement more verbose, but also more clear and less magical e.g.

```clojure
;; match value directly
(is (match? {:a (within-delta 0.1 3)}
            {:a 3.05}))

;; match nested values
(is (match? (match-with [number? (within-delta 0.1)]
                        {:a 3)}
            {:a 3.05}))

```

src/cljc/matcher_combinators/utils.cljc
```diff
@@ -1,9 +1,17 @@
-(ns matcher-combinators.utils)
+(ns matcher-combinators.utils
+  "Internal use only. Subject (and likely) to change.")
 
-(defn roughly? [expected actual delta]
-  (and (number? actual)
-       (>= expected (- actual delta))
-       (<= expected (+ actual delta))))
+(defn processable-number? [v]
```
fernando-nubank, line 4: I'm not sure if we should hide this exception. If I'm trying to match a number against something invalid I would expect an exception.

fernando-nubank:
> This would make the match-roughly? replacement more verbose, but also more clear and less magical

I prefer to be more explicit and less magical :)

dchelimsky:
> > This would make the match-roughly? replacement more verbose, but also more clear and less magical
> 
> I prefer to be more explicit and less magical :)

as ordens

src/cljc/matcher_combinators/utils.cljc
dchelimsky: This will result assertion failure (a mismatch), but not an exception.

We were already returning false for non-numeric values. This just treats ##Inf, ##-Inf, and ##NaN as non-numeric for the purposes of this matcher. If the code produces ##Inf, that will show up in the failed assertion, which is sufficient info IMO. WDYT?

fernando-nubank: OK, makes sense now

PR #140, opened 2020-07-01
refactor processable-number? helper fn for within-delta

dchelimsky:
- check for decimal instead of catching Exception when asking if
  decimal value is infinite or NaN
- update test to generate more types of numbers for delta

This is a minor internal refactoring that will not effect behavior, so not even mentioning it in changelog or bumping version. It'll get subsumed by the next release.


dchelimsky:
FYI @lucasdf

PR #153, opened 2020-09-30
Add linter check to github PRs + run linter

philomates:


philomates:
cljfmt doesn't support parsing `##Inf` so the linter crashes and the step fails. I think I'll just leave this out for now

PR #157, opened 2021-05-31
add flare string diffing to output

philomates:
Midje, when reporting test failures where the expected and actual are strings, uses https://github.com/duckyuck/flare to do string diff'ing.

for example you might see output like
```
Expected:
"hello world!"
Actual:
"hello world"
Diffs: strings have 1 difference (91% similarity)
                expected: "... world(!)"
                actual:   "... world(-)"
```

This PR is a proposal to add this to matcher-combinators.

For something like this
```clojure
(fact 
  ["abc" "ayz"] => (match (matchers/in-any-order ["ay*" "ab*"])))
```

Before:
```
FAIL at (scratch.clj:50)
Actual result:
["abc" "ayz"]
Checking function: (match (matchers/in-any-order ["ay*" "ab*"]))
        [:mismatch
 ((mismatch "ab*" "abc")
  (mismatch "ay*" "ayz"))]
```

after
```
FAIL at (scratch.clj:50)
Actual result:
["abc" "ayz"]
Checking function: (match (matchers/in-any-order ["ay*" "ab*"]))
        [:mismatch
 ((mismatch "ab(*)" "ab(c)")
  (mismatch "ay(*)" "ay(z)"))]
```

or for

```clojure
(deftest bar
  (is (match? "hello o brave new world!"
              "hello world!")))
```
before:
```
FAIL in (bar) (test_test.clj:120)
expected: (match? "hello o brave new world!" "hello world!")
  actual: (mismatch
 "hello o brave new world!"
 "hello world!")
```
after
```
FAIL in (bar) (test_test.clj:120)
expected: (match? "hello o brave new world!" "hello world!")
  actual: (mismatch
 "hello (o brave new )world!"
 "hello (------------)world!")
```

Any suggestions on making it clearer what is expected, actual, and the diff between them?

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -14,7 +14,8 @@
                      provided [[:inner 0]]
                      tabular  [[:inner 0]]}}
 
-  :dependencies [[org.clojure/clojure "1.10.1"]
+  :dependencies [[flare "0.2.9" :exclusions [org.clojure/clojure]]
+                 [org.clojure/clojure "1.10.1"]
                  [org.clojure/spec.alpha "0.2.187"]
                  [org.clojure/math.combinatorics "0.1.6"]
                  [midje "1.9.9" :exclusions [org.clojure/clojure]]]
--- a/src/clj/matcher_combinators/clj_test.clj
+++ b/src/clj/matcher_combinators/clj_test.clj
@@ -88,7 +88,7 @@
        (core/matcher? ~matcher)
        (let [result# (core/match
                       (matchers/match-with ~type->matcher ~matcher)
-                      ~actual)]
+                       ~actual)]
          (clojure.test/do-report
           (if (core/indicates-match? result#)
             {:type     :pass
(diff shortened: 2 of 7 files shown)


philomates:
> 1. It applies when using regex as well?

It should only apply when both the expected and actual are strings

> 2. Could we add a way to turn it off? I'm wondering if may affect performance when dealing with big maps with lot's of strings. wdyt?

Good point. So the diff code only gets run after a mismatch is detected, so it will only slow things down if the expected/actual strings are large. I tried it on a sorta large string and it took 2 seconds to print, which isn't great

To be honest I don't really know if this change is worth the effort. I find the output kind of confusing.
Maybe leaving it as off by default and allowing people to turn it on if they need it? The user that asked for this wanted to compare large string files, which I don't think is so common

gabrielgiussi:
> leaving it as off by default and allowing people to turn it on if they need it

I like this idea if is not very complex to add.

Or could we even think in a way to plug-in matchers? So this particular engineer could add the dependency to https://github.com/duckyuck/flare but it won't be included in matcher-combinators.
This may be harder though, I really don't know much about the internals of matcher-combinators so it is hard for me to estimate the effort such thing requires.

Issue #99, opened 2020-02-28
Support matching negating an item in a Vector


markomafs:
### Needs 
We have some cases where I want to match that an item is not present in a vector: 

i.e.: 
```clojure
(is (match? 
         (m/embeds [map1 (dont-want-this map2)]) 
         [map1 map2 map3]))
```
I didn't find it. 
Is there a way to do that?

philomates:
So there isn't currently a way to do this.
I'm a little afraid that such behavior wouldn't be used often and might be a bit confusing, but this is mostly a feeling I have.

Reflecting on this a little, I wonder if it goes a bit against the declarative focus of matcher-combinators. Like usually with matcher-combinators you say "the data should look like this", and here you are saying it "shouldn't look like this". I don't know if this is an important distinction or how it plays out, but I guess it is interesting to reflect on. Also note that we already have `absent` in the context of specifying that a key in a map is absent, so it isn't like we're adhering to the constructionist declarative approach fully at the moment.

markomafs:
Agree. 
I think we could approach this by discussing conceptually how matcher-combinators will position itself on this matter and the result would be: 
* we won’t support negative behavior and properly document it. 
* we will support and implement it. 


I’m available to discuss that.

philomates:
https://github.com/nubank/matcher-combinators/pull/158 should implement this. 
To assert an entry doesn't appear in a vector you can now do
```
(testing "assert everything in the vector is an even number via negation (generally avoid this)"
  (is (match? (mismatch (embeds [odd?])) actual)))
```

PR #158, opened 2021-06-01
implement a "mismatch" matcher

philomates:
useful for asserting that something is not in the result: https://github.com/nubank/matcher-combinators/issues/99
for example:
```clojure
(deftest mismatch
  (testing "predicate not-match"
    (is (match? [1 (m/mismatch odd?) 3]
                [1 2 3])))
  (testing "declarative not-match"
    (is (match? [1 (m/mismatch {:a 1}) 3]
                [1 {:a 2 :b 1} 3])))
  (testing "in-any-order with not-match"
    (is (match? (m/in-any-order
                 [odd? pos? (m/mismatch odd?)])
                [1 2 3]))))
```

### open questions:

- does using `mismatch` lead to more readable tests in general?
- does the new expressive power out-weigh the mental burden of running ` semantics` in your head?

for both of these questions I decided to discourage negation matching in the [README](https://github.com/nubank/matcher-combinators/pull/158/files#r646786051) and the [docstring](https://github.com/nubank/matcher-combinators/pull/158/files#diff-753006ace639d244b0bc581f46c36b9f61d337f1f4c1fac2e7e19ac4dbd4ef3cR91-R92)

### example failure messages
```
FAIL in (absent-matcher) (test_test.clj:124)
expected: (match? [1 (m/mismatch {:a 2}) 3] [1 {:a 2} 3])
  actual: [1
 (mismatch "expected mismatch from: {:a 2}" {:a 2})
 3]


FAIL in (absent-pred) (test_test.clj:120)
expected: (match? [1 (m/mismatch even?) 3] [1 2 3])
  actual: [1
 (mismatch
  "expected mismatch from: clojure.core$even_QMARK_@6a0a7d89"
  2)
 3]
```

mismatch

### other
importantly this doesn't conflate exceptions and values like the Midje `=not=>` does, which has traditionally lead to many uncaught bugs in our test suites:

```clojure
(facts "=not=> vs mismatch"
  (fact "wHHYYY does this pass?"
    (throw (ex-info "boom" {})) =not=> 1)
  (fact "we sleep well at night knowing this doesn't pass"
    (throw (ex-info "boom" {})) => (match (m/mismatch 1))))
```


src/cljc/matcher_combinators/matchers.cljc
```diff
@@ -74,6 +74,12 @@
   [pred]
   (core/->PredMatcher pred (str "predicate: " pred)))
 
+(defn not-matcher
```
philomates, line 77: is `not` a better name? or can someone think of something nicer?

dchelimsky: I like `not` or `mismatch` (since that's what's in the error message)

src/cljc/matcher_combinators/core.cljc
```diff
@@ -460,15 +476,43 @@
...
+  Matcher
+  (-matcher-for [this] this)
+  (-matcher-for [this _] this)
+  (-match [this actual]
+    (let [{::result/keys [type weight] :as result} (match expected actual)]
+      (if (= :match type)
+        {::result/type   :mismatch
+         ::result/value  (model/->ExpectedMismatch
+                          (printable-matcher expected)
+                          actual)
```
philomates, line 501: this is needed to be able to print a mismatch message that includes the original matcher when a mismatch is expected. Like the `(embeds [odd?])` in the example below
```
FAIL in (absent-matcher) (matchers_test.clj:386)
assert an entry is definitely not in a sequence
expected: (match? (m/mismatch (m/embeds [odd?])) [1 3 5 7])
  actual: (mismatch
 expected mismatch from:
 (embeds [#function[clojure.core/odd?]])
 [1 3 5 7])
```

dchelimsky (approved):

src/cljc/matcher_combinators/core.cljc
```diff
@@ -143,6 +144,12 @@
      ::result/weight 1})
   (-name [_] 'absent))
 
+(defmethod clojure.pprint/simple-dispatch Absent [absent]
+  (pr (-name absent)))
```
philomates, line 148: this should technically go in the `printer` namespace but it would make things a bit more messy in terms of dependencies / using `-name` outside of `core` (which I guess is fine, so I'm okay with pushback to move it there)

anyways before this
```
FAIL in (mismatcher-matcher) (matchers_test.clj:384)
assert presence of key via double negation
expected: (match? (m/mismatch {:a m/absent}) {:b 1})
  actual: (mismatch
 expected mismatch from:
 {:a {}}
 {:b 1})

```

and after

```
FAIL in (mismatcher-matcher) (matchers_test.clj:384)
assert presence of key via double negation
expected: (match? (m/mismatch {:a m/absent}) {:b 1})
  actual: (mismatch
 expected mismatch from:
 {:a absent}
 {:b 1})
```

src/cljc/matcher_combinators/core.cljc
```diff
@@ -479,6 +486,32 @@
        ::result/weight 1}))
   (-name [_] 'predicate))
 
+(defn- printable-matcher [matcher]
+  (try
+    (if-let [n (-name matcher)]
+      `(~(symbol n) ~(:expected matcher))
+      matcher)
+    (catch IllegalArgumentException _e
```
philomates, line 494: is this the best way to see if `-name` is implemented or not?

README.md
```diff
@@ -176,17 +176,34 @@ for a specific value, e.g.
...
+
+```clojure
+(deftest avoid-negative-matchers
+  (testing "normal assertion that `:a` is present"
+    (match? {:a any?}
+            actual))
+  (testing "double negation version"
+    (match? (matcher-combinators.matchers/mismatch {:a matcher-combinators.matchers/absent})
+            actual)))
+```
```
philomates, line 200: how does this look/read?

src/cljc/matcher_combinators/core.cljc
```diff
@@ -12,7 +13,7 @@
     [expected]
     [expected t->m]
     "Do not call directly. Implementation for matcher-combinators.matchers/matcher-for.")
-  (-name [this]
+  (-base-name [this]
```
philomates, line 16: `-name` shadows something in clojurescript which results in some build warning

README.md
dchelimsky: I can't make a suggestion because it wraps triple "`" in triple "`" :/

How about "are generally discouraged due to the adverse effect they can have on code readability."?

Also, I think we should recommend `:a-key nil` instead of `:a-key absent` except where the actual presence of the key really matters (which is almost never IMO).

PR #160, opened 2021-06-07
Lift matcher name logic to matcher protocol

philomates:
generally a nice little standardization of existing code but also needed to implement printing underlying matcher for the mismatch matcher: https://github.com/nubank/matcher-combinators/pull/158/files#r646551305

### DIFF
--- a/src/cljc/matcher_combinators/matchers.cljc
+++ b/src/cljc/matcher_combinators/matchers.cljc
@@ -66,13 +66,14 @@
   (core/->Regex expected))
 
 (def absent
-  "Value-position matcher for maps that matches when containing map doesn't have the key pointing to this matcher."
+  "Value-position matcher for maps that matches when containing map doesn't
+  have the key pointing to this matcher."
   (core/->Absent))
 
 (defn pred
   "Matcher that will match when `pred` of the actual value returns true."
   [pred]
-  (core/->PredMatcher pred (str "predicate: " pred)))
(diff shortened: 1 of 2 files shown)


src/cljc/matcher_combinators/matchers.cljc
```diff
@@ -66,13 +66,14 @@
...
-  "Value-position matcher for maps that matches when containing map doesn't have the key pointing to this matcher."
+  "Value-position matcher for maps that matches when containing map doesn't
+  have the key pointing to this matcher."
   (core/->Absent))
 
 (defn pred
   "Matcher that will match when `pred` of the actual value returns true."
   [pred]
-  (core/->PredMatcher pred (str "predicate: " pred)))
+  (core/->PredMatcher pred `(~'pred ~pred)))
```
philomates, line 76: sneaking this little improvement in as well:

before this change 
```
    (is (match? even? 1)))
```
would print
```
FAIL in (absent-matcher) (matchers_test.clj:384)
assert an entry is definitely not in a sequence
expected: (match? even? 1)
  actual: (mismatch "clojure.core$even_QMARK_@9d37f5c" 1)
```
and after it prints
```
FAIL in (absent-matcher) (matchers_test.clj:384)
assert an entry is definitely not in a sequence
expected: (match? even? 1)
  actual: (mismatch (pred #function[clojure.core/even?]) 1)
```

dchelimsky (approved):

PR #162, opened 2021-06-23
Fixed `printable-matcher` cljs error catch

Aeonax:
Don't know what is going on here, but looks like there was a misspell which sent me a lot of warnings in the console.

```
------ WARNING #1 - :undeclared-var --------------------------------------------
 Resource: matcher_combinators/core.cljc:491:51
 Use of undeclared Var matcher-combinators.core/Exception
```


philomates (approved):
thanks for the patch! I'm not using clojurescript so it is helpful to receive these reports and patches from those that do :bow:

philomates:
deployed as `3.2.1`

PR #167, opened 2022-02-21
rename `matcher-combinators.utils/abs` to avoid shadowing `clojure.core/abs`

philomates:
I'm seeing the following warning a lot in my REPL:
```
WARNING: abs already refers to: #'clojure.core/abs in namespace: matcher-combinators.utils, being replaced by: #'matcher-combinators.utils/abs
```

I think this PR should address it


philomates:
covered in https://github.com/nubank/matcher-combinators/pull/169

PR #169, opened 2022-03-28
upgrade to clojure-1.11.0, eliminate "already refers to" warnings

dchelimsky:
Clojure-1.11.0 introduces some new functions to clojure.core that replace custom functions in some libraries, which produces "WARNING: xxx already refers to" warnings.

This PR eliminates those warnings by upgrading a transitive dependency that avoids it and using :refer-clojure with `:exclude` for warnings emanating from this lib.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -14,10 +14,13 @@
                      provided [[:inner 0]]
                      tabular  [[:inner 0]]}}
 
-  :dependencies [[org.clojure/clojure "1.10.3"]
-                 [org.clojure/spec.alpha "0.2.187"]
+  :dependencies [[org.clojure/clojure "1.11.0"]
+                 [org.clojure/spec.alpha "0.3.218"]
                  [org.clojure/math.combinatorics "0.1.6"]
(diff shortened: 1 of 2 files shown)


philomates (approved):
nice, thanks! I'll close https://github.com/nubank/matcher-combinators/pull/167 then because it doesn't do the `resolve` trick

project.clj
```diff
@@ -14,10 +14,13 @@
...
-  :dependencies [[org.clojure/clojure "1.10.3"]
-                 [org.clojure/spec.alpha "0.2.187"]
+  :dependencies [[org.clojure/clojure "1.11.0"]
+                 [org.clojure/spec.alpha "0.3.218"]
                  [org.clojure/math.combinatorics "0.1.6"]
-                 [midje "1.10.4" :exclusions [org.clojure/clojure]]]
+                 [midje "1.10.5" :exclusions [org.clojure/clojure]]
+                 ;; override midje's dependency on an old version
+                 ;; of pretty
+                 [io.aviso/pretty "1.1.1"]]
```
philomates, line 23: https://github.com/marick/Midje/pull/480

PR #171, opened 2022-04-08
support map-like values as actuals in embeds matcher

dchelimsky:
# Problem

This fails

```clojure
(deftest person-test
  ,,,
  (match? {:first-name "David"} (fn-that-returns-datomic-entity-including {:first-name "David"}))
```

The reason is that the `embeds` matcher checks to see that the actual value is a `map?`, which a Datomic EntityMap is not.

# Solution

Generalize the type check for the actual value to allow for anything that is `associative?` but not `sequential?` (to avoid comparing vectors with maps).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ All notable changes to this project will be documented in this file. This
 change log follows the conventions of
 [keepachangelog.com](http://keepachangelog.com/).
 
+## [3.5.0]
+- embeds matcher supports map like (associative, not sequential) actual values
+
 ## [3.4.0]
 - Eliminate "already refers to" warning on `abs` for clojure-1.11.0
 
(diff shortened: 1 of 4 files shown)


marcobiscaro2112 (approved):

philomates:
> IMO equals should not match for different types, but I could be convinced.

sounds good, I think I agree. Wanted to see what folks thought though because it did cause a moment of confusion for me when I wrapped it in a `m/equals` to try to assert all keys were accounted for. Maybe a case for the `via` thing.

> a generative test that produces any data structure besides a map and always fails to match

wouldn't the following cover it?
```clojure
(is (not (core/indicates-match?
           (core/match (matchers/equals {:a 1}) map-like))))
```

dchelimsky:
> wouldn't the following cover it?

Yes! Have at it (or I can do it, but not today).

PR #174, opened 2022-09-07
move midje to scope provided

dchelimsky:
# Problem

Users who don't rely on midje support incur a transitive dependency on midje anyway.

# Solution

Move midje dependency to :scope "provided"


src/clj/matcher_combinators/midje.clj
```diff
@@ -89,7 +96,9 @@
       (throw (ArityException. arg-count "throws-match")))))
 
 (checkers.defining/defchecker throws-match
```
marcobiscaro2112, line 98: Does `defchecker` support metadata? Can we/do we need to add `^:deprecated` here too?

src/clj/matcher_combinators/midje.clj
```diff
@@ -130,11 +139,15 @@
          ::result/weight 1}))))
 
 (def match-equals
```
marcobiscaro2112, line 141: Add `^:deprecated`?

marcobiscaro2112 (approved):

src/clj/matcher_combinators/midje.clj
dchelimsky: Done here and for `match-roughly`

src/clj/matcher_combinators/midje.clj
dchelimsky: Yes, as it turns out. Done.

.clj-kondo/marick/midje/config.edn
```diff
@@ -0,0 +1,80 @@
+{:hooks {:analyze-call {midje.sweet/tabular marick.midje/tabular}}
```
philomates, line 1: @borkdude do you know if committing these configs here in combination with tagging `midje` as `:scope "provided"` in the `matcher-combinators` deps mean that the `midje` configs for `clj-kondo` would be still be picked up via `clj-kondo --lint $(clojure -Spath) --dependencies --copy-configs` ?

Like does `--copy-configs` copy all `.clj-kondo` configs from a lib even if it is from a different org namespace?

CHANGELOG.md
```diff
@@ -3,6 +3,12 @@ All notable changes to this project will be documented in this file. This
 change log follows the conventions of
 [keepachangelog.com](http://keepachangelog.com/).
 
+## [3.6.1]
+- deprecate the matcher-combinators.midje ns and functions in it
```
philomates, line 7: Fun to imagine this resulting in tens of thousands of new "deprecated" lint warnings in the test suites that Nubank still hasn't migrated away from Midje. Maybe this will help stoke the fire :P

philomates (approved):

philomates:
maybe (too) dirty, but we could also move all the midje-related matcher-combinator stuff into midje itself, but still under the same `matcher-combinator.midje` namespace so that any midje users would see the same stuff/not require any code changes, but we could go clean from `midje` here

scottbale:
Documenting here for clarity: Newer PR #197 deleted `project.clj` (which was where "scope provided" was specified). In its place, `deps.edn` confines `midge` dependency to the `:extra-deps` of `:dev` alias. (See also `:midje-test` alias.) I will rename this PR to make that more clear.

CHANGELOG.md
dchelimsky: Nobody cares about deprecation warnings except @acamargo and me.

CHANGELOG.md
```diff
@@ -3,6 +3,12 @@ All notable changes to this project will be documented in this file. This
 change log follows the conventions of
 [keepachangelog.com](http://keepachangelog.com/).
 
+## 3.8.4 / 2023-02-08
+- deprecate the matcher-combinators.midje ns and functions in it
+- move midje dependency to :scope "provided"
+  - this eliminates the transitive dep for those not using midje features
```
scottbale, line 9: These two lines are no longer valid. #197 (since merged into this PR) removed `project.clj` and effectively already removed midje as a transitive dependency.

dchelimsky:
@philomates I updated the CHANGELOG and README. Please re-review. If it's good for you I'll go ahead and merge it.

scottbale (approved):

Issue #188, opened 2022-12-07
Consider adding a deps.edn


lread:
## Currently
matcher-combinators only has a `project.clj`

## Opportunity
Adding a `deps.edn` allows for easy:
- access to unreleased works-in-progress to support testing via git dep
- access to works-in-progress locally via `:local\root` dep

## Next steps
If you think this is a good idea, happy to help with a PR.

## Related
#174 - I would think a `deps.edn` would not reference midje.

philomates:
thanks for the suggestion, excited to get all the goodness of the `deps.edn` tooling.
Addressed in https://github.com/nubank/matcher-combinators/pull/197

Issue #156, opened 2021-02-09
match? seems broken in CLJS


wilkerlucio:
Hello, I'm trying to use it to test with clojure.test in CLJS, but the `match?` always returns true, example case:

```clojure
(deftest sample-test
  (is (match? (m/embeds {:foo "bar"}) {})))
```

The test passes, I expect it to fail.

philomates:
Hey @wilkerlucio!
I'm pretty unfamiliar with cljs setups, but I copied your example into `test/cljs/matcher_combinators/cljs_example_test.cljs` and ran it both with `lein test-node` and `lein test-phantom` and in both cases it fails. 
In what context are you running the example?

wilkerlucio:
Hello @philomates , thanks for checking!

The setup I'm trying is with Shadow CLJS on the browser, I made a complete repro demo, can you try this? https://github.com/wilkerlucio/match-comb-shadow-repro

philomates:
nice, I got it working (err, that is passing incorrectly)
Some notes before I log off for the day:
 - expected failure behavior is demonstrated when you target nodejs via `shadow-cljs compile test-λ && node out/node-tests.js` after adding the following to shadow-cljs.edn
 ```
          :test-λ {:target    :node-test
                   :output-to "out/node-tests.js"
                   :ns-regexp "-test$"
                   :autorun   true}
```
 - shadow-cljs has some weird monkey-patching of cljs.test https://github.com/thheller/shadow-cljs/blob/master/src/main/shadow/test.cljs. Found this thanks to a comment in https://github.com/lambdaisland/chui/blob/master/modules/chui-core/src/lambdaisland/chui/runner.cljs#L85-L87
 - matcher-combinators does some weird stuff to register `match?` in cljs: https://github.com/nubank/matcher-combinators/blob/master/src/cljc/matcher_combinators/cljs_test.cljc#L15-L25

I'm suspecting that the test reporting for `match?` isn't properly being handled in when targeting `:browser-test`.
I'll try to dig a little farther tomorrow but maybe @plexus knows right off the bat what's going on here, given the work with chui?

plexus:
[...]

Issue #83, opened 2019-08-19
Not registering failures on async tests


mauricioszabo:
I'm using Shadow-CLJS, and devcards, to test async code. If I try to use this library to test it, and there's a failure, it's not being registered. It's kinda tedious to reproduce this error, so I created a repository to reproduce it: https://github.com/mauricioszabo/bug-info-example

To simplify, a test with devcards like this one:

```clojure
(defn some-code [a]
  (js/Promise. (fn [resolve]
                 (js/setTimeout #(resolve (+ a 10))
                                1000))))

(set! cards/test-timeout 8000)
(cards/deftest some-async-test
  (async done
    (.. (some-code 12)
        (then (fn [res] (is (match? 21 res))))
        (then done)
        (catch done))))

(cards/deftest correct-async-test
  (async done
    (.. (some-code 12)
        (then (fn [res] (is (= 21 res))))
        (then done)
        (catch done))))
```

Will produce the following HTML output:

![image](https://user-images.githubusercontent.com/138037/63275822-9a2cc480-c278-11e9-8d9b-8a83bd5950ce.png)

On the test with `matcher-combinators`, neither the failure nor the assertion is registered.

kommen:
I'm running into the same problem: mismatches using `match?` are not registered and when using `async` in cljs tests. 

Looking at the report handlers, it looks like it is handled the same way as the `:fail` handlers in cljs.test, so it's not obvious to me what is wrong here:

https://github.com/nubank/matcher-combinators/blob/638b7c8dd0ceb881f4edcb38676c3b60ad0f6a10/src/cljc/matcher_combinators/cljs_test.cljc#L103-L121

plexus:
The async is a red herring here, matcher-combinators does not work with devcards or kaocha-cljs in general:

``` clojure
(cards/deftest non-async-test
  (is (match? 22 12)))
```

![2019-09-11-145601_1555x437_scrot](https://user-images.githubusercontent.com/32212/64699060-4e9abe80-d4a4-11e9-9d68-5109a83a668a.png)

Because neither of these test runners have implementations for the `:matcher-combinators/mismatch` message. The situation is similar for Clojure, but there Kaocha explicitly contains [support for matcher combinators](https://github.com/lambdaisland/kaocha/blob/v0.0-529/src/kaocha/matcher_combinators.clj), I have written about this a little [over here](https://cljdoc.org/d/lambdaisland/kaocha/0.0-529/api/kaocha.report).

[...]

PR #189, opened 2022-12-13
Make cljs version work with browser tests

philomates:
Should address https://github.com/nubank/matcher-combinators/issues/156 and https://github.com/nubank/matcher-combinators/issues/83, which @zampino and I ran into today ourselves.
The idea behind the fix is to apply the changes done in https://github.com/nubank/matcher-combinators/commit/7635d040fb81ad44b3587644d75bba8961269dc7 for clj clojure.test to the cljs branch of the code.

note that I still haven't figured out how to get ansi color escape codes to print well in the browser. If anyone has ideas please chime in. (EDIT: calling `(matcher-combinators.ansi-color/disable!)
` from https://github.com/nubank/matcher-combinators/pull/196 can do it)
20221213_17h25m12s_grim

I'm adding a `deps.edn` file for ease of launching `shadow-cljs`. I imagine nubank will still do deploys using its `lein` based-infra (assuming a migration over to clj deps hasn't happened yet) but probably still nice to have `deps.edn` as mentioned also by @lread in https://github.com/nubank/matcher-combinators/issues/188

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -1,0 +1,16 @@
+{:paths ["src/clj" "src/cljc" "src/cljs"]
+ :deps {org.clojure/clojure {:mvn/version "1.11.1"}
+        org.clojure/spec.alpha {:mvn/version "0.3.218"}
+        org.clojure/math.combinatorics {:mvn/version "0.1.6"}}
+
+ :aliases
+ {:dev
+  {:extra-paths ["dev" "test/cljc" "test/cljs" "test/clj"]
+   :extra-deps {org.clojure/test.check {:mvn/version "1.1.1"}
+                orchestra/orchestra {:mvn/version "2021.01.01-1"}}}
+
+  :cljs-test
+  {:extra-paths ["test/cljc" "test/cljs"]
+   :main-opts ["-m" "shadow.cljs.devtools.cli"]
+   :extra-deps {org.clojure/test.check {:mvn/version "1.1.1"}
+                thheller/shadow-cljs {:mvn/version "2.18.0"}}}}}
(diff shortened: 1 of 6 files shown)


src/cljc/matcher_combinators/cljs_test.cljc
```diff
@@ -9,6 +9,18 @@
...
+(defrecord CljsMismatch [summary match-result])
+
+(defn tagged-for-pretty-printing [actual-summary result]
+  (->CljsMismatch actual-summary result))
+
+#?(:cljs
+   (extend-protocol IPrintWithWriter
+     CljsMismatch
+     (-pr-writer [cljs-mismatch writer _]
+       (-write writer (printer/as-string (-> cljs-mismatch :match-result ::result/value))))))
```
philomates, line 21: is there a way to do this without going through `printer/as-string`? Something similar to the clj version that does:
```clojure
(defmethod clojure.core/print-method ::mismatch [{:keys [match-result]} out]
  (binding [*out* out]
    (printer/pretty-print (::result/value match-result))))
```

deps.edn
```diff
@@ -0,0 +1,16 @@
+{:paths ["src/clj" "src/cljc" "src/cljs"]
```
dchelimsky, line 1: Maintaining both deps.edn and project.clj is error prone. WDYT about just moving the whole project to tools.deps (deps.edn)?

philomates: I like the idea but it would require somebody at Nubank to migrate the lib release automation, right?

philomates: ah right, then yeah I can try to do that

Issue #149, opened 2020-09-14
`seq-of` matcher


dchelimsky:
Add a matcher that matches a non-empty sequence of elements that match a supplied matcher, e.g.

```clojure
(match? (seq-of {:name string? :id uuid?})
        [{:name "Michael"
         :id    #uuid "c70e35eb-9eb6-4e3d-b5da-1f7f80932db9"}])
```

This impl works, but the resulting failure message leaves something to be desired:

```clojure
(defn seq-of [matcher] (fn [coll] (and (seq coll) (every? (partial matcher-combinators.standalone/match? matcher) coll))))
```

philomates:
giving it a shot in https://github.com/nubank/matcher-combinators/pull/191, what do you think?

Issue #18, opened 2018-02-12
Implement matcher for checking a property against some or all of the sequence values


philomates:
similar to midje's [`has`](https://github.com/marick/Midje/wiki/Checking-sequential-collections#has-checks-a-property-against-some-or-all-of-the-sequence-values)

This might not fit well with the approach that `matcher-combinators` takes, depending on the actual data-structure. Could potentially handle this with infinite series `(match (repeat odd?) [1 3 5])`

dchelimsky:
@philomates doesn't `(embeds [m1 m2 ,,,] <sequence>)` cover this? Or am I missing something?

philomates:
yeah, I think so. This wasn't really a fleshed out idea, just an observation by a colleague that I turned into an issue. Going to close it out

PR #191, opened 2022-12-14
Add `seq-of`, `all-of`, and `any-of` combinators

philomates:
addresses https://github.com/nubank/matcher-combinators/issues/149

an example of the difference between the new `seq-of` matcher and @dchelimsky 's temporary workaround `(defn seq-of [matcher] (fn [coll] (and (seq coll) (every? (partial matcher-combinators.standalone/match? matcher) coll))))`
20221214_13h57m17s_grim

not sure if this additional expressiveness is opening a can of worms in terms of the complexity available to users. It has been asked for before though in https://github.com/nubank/matcher-combinators/issues/18 and https://github.com/nubank/matcher-combinators/issues/149 and a few times during in-person chats

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,13 +4,9 @@ change log follows the conventions of
 [keepachangelog.com](http://keepachangelog.com/).
 
 ## Unreleased
-- Introduce `seq-of`, `any-of`, and `all-of` matchers combinators
-   * `seq-of` takes an expected matcher and creates a new matcher over a
-     sequence, where each element matches the provided expected matcher.
-   * `any-of` takes any number of matchers, successfully matching when at least
-     one matches.
-   * `all-of` takes any number of matchers, successfully matching when all
-     match.
+- Add `seq-of` matcher, which takes a matcher, successfully matching when each element matches the provided matcher.
+* Add `any-of` matcher, which takes any number of matchers, successfully matching when at least one matches.
+* Add `all-of` matcher, which takes any number of matchers, successfully matching when all match.
 
 - Add 2-arity `pred` matcher where the second argument is a description text.
   Useful for mismatch messages when the pred is an anonymous function.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ change log follows the conventions of
 [keepachangelog.com](http://keepachangelog.com/).
 
 ## Unreleased
+- Introduce `seq-of`, `any-of`, and `all-of` matchers combinators
+   * `seq-of` takes an expected matcher and creates a new matcher over a
+     sequence, where each element matches the provided expected matcher.
+   * `any-of` takes any number of matchers, successfully matching when at least
+     one matches.
+   * `all-of` takes any number of matchers, successfully matching when all
+     match.
+
 - Add 2-arity `pred` matcher where the second argument is a description text.
   Useful for mismatch messages when the pred is an anonymous function.
 - Allow globally configuring ANSI color emission via newly added `enable!` and
--- a/src/cljc/matcher_combinators/core.cljc
+++ b/src/cljc/matcher_combinators/core.cljc
@@ -283,27 +283,23 @@
        actuals])))
 
 (defn- sequence-match [expected actual subseq?]
-  (if-not (sequential? actual)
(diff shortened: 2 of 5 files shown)


src/cljc/matcher_combinators/matchers.cljc
```diff
@@ -114,6 +120,11 @@
   [transform-actual-fn expected]
   (core/->ViaMatcher transform-actual-fn expected))
 
+(defn either
+  "A matcher that successfully matches if one of the two provided matchers matches."
+  [left-expected right-expected]
+  (core/->Either left-expected right-expected))
```
philomates, line 126: I wonder if `either` should take in an arbitrary number of arguments instead of just 2.

It starts to look like `clojure.core/some-fn`, which makes me wonder if we should also introduce an analog to `clojure.core/every-pred`, say `m/every`, so you can do things like

```clojure
(is 
  (match?
    (m/every (m/prefix [1 2 3]) #(> (count %) 5))
    (some-expr))
  "a list starting with 1 2 3 that has more than 5 elems")
```

ericdallo: Looks a little bit more complex and may cause some confusion for users, I suggest sticking with either or even drop it for now, fine with keeping it though.

dchelimsky: We've got two separate problems here:
- apply a single matcher to every element in a sequence
- apply n matchers to a single element

`either` definitely leaves me wanting something like the `m/every` that you propose, so, for me, the pair of `seq-of` and `every` seem good from a functionality perspective. We should explore the names a bit more.

dchelimsky: How about `any` instead of `Either`? I'd prefer to avoid anything that implies monads :)

philomates: I'm fine with `any`; `either` to me connotes two-ness ("either one or the other"). 
note that `any` has the downside of potentially being conflated with clojure's `(defn any? [x] true)` which is pretty different

dchelimsky: `any-of` and `all-of`?

philomates: done in https://github.com/nubank/matcher-combinators/pull/191/commits/c38bd9c6611da087ba8b5c0f60fd455310924c63

src/cljc/matcher_combinators/core.cljc
```diff
@@ -503,6 +513,49 @@
...
+    (reduce (fn [_acc matcher]
+              (if (= ::end matcher)
+                {::result/type   :match
+                 ::result/value  actual
+                 ::result/weight 0}
+                (let [result (match matcher actual)]
+                  (when-not (indicates-match? result)
+                    (reduced
+                      {::result/type   :mismatch
+                       ::result/value  (model/->Mismatch (concat ['all-of] matchers) actual)
```
philomates, line 552: note that mismatch info doesn't tell you want matched and what didn't. I imagine we could improve the info returned, but maybe would be nice to in a follow-up(?)

CHANGELOG.md
```diff
@@ -4,6 +4,14 @@ change log follows the conventions of
...
 
 ## Unreleased
+- Introduce `seq-of`, `any-of`, and `all-of` matchers combinators
+   * `seq-of` takes an expected matcher and creates a new matcher over a
+     sequence, where each element matches the provided expected matcher.
+   * `any-of` takes any number of matchers, successfully matching when at least
+     one matches.
+   * `all-of` takes any number of matchers, successfully matching when all
+     match.
+
```
dchelimsky, line 14:
```suggestion
- Add `seq-of` matcher, which takes a matcher, successfully matching when each element matches the provided matcher.
* Add `any-of` matcher, which takes any number of matchers, successfully matching when at least one matches.
* Add `all-of` matcher, which takes any number of matchers, successfully matching when all match.

```

dchelimsky:
OK - this with the other 2 make for a solid release. I'll merge this and release all of this by tomorrow - maybe today.

dchelimsky:
Released in 3.8.0: https://clojars.org/nubank/matcher-combinators

Issue #193, opened 2023-01-20
Question: Minimum supported Clojure version for matcher-combinators?


lread:
Hiya, I continue to really enjoy using matcher-combinators, thanks again for an awesome library!

One thing I'm not entirely confident of is the minimum Clojure version required to run matcher-combinators. 

I see that [the current pom](https://repo.clojars.org/nubank/matcher-combinators/3.7.2/matcher-combinators-3.7.2.pom) specifies Clojure 1.11.1:

```xml
    <dependency>
      <groupId>org.clojure</groupId>
      <artifactId>clojure</artifactId>
      <version>1.11.1</version>
    </dependency>
```
And that strongly implies that Clojure 1.11.1 is the minimum recommended version.

If that's the case that's fine, but maybe we should also state it clearly in the README?
Lemme know, can do a quick PR for that.

Why do I ask, you ask? Well, I'm doing some maintenance on pomegranate, and I thought, [woo matcher-combinators would be nice in these tests](https://github.com/clj-commons/pomegranate/issues/152), but then I remembered, oh yeah, pomegranate wants to support back at least Clojure 1.8. And I want to test pomegranate with each supported Clojure version on CI.

If 1.11.1 is not the actual min supported version for matcher-combinators,
and you want to add CI tests against older versions of Clojure, I am happy to help with that.

philomates:
We have a clojure 1.8 profile in `project.clj` but running `lein with-profile 1.8 test` errors due to `clojure.spec.alpha`, which I guess doesn't work on 1.8(?).
Spec isn't super used in matcher-combinators so maybe there is a way around it. But we do use the namespace keyword alias thing (`::`), which is clojure 1.9, right?

lread:
Thanks for the reply @philomates!

I think the `::` syntax has been there since the start:

```Clojure
❯ clj -Sdeps '{:deps {org.clojure/clojure {:mvn/version "1.0.0"}}}'
Clojure 1.0.0-
user=> ::foo
:user/foo
user=> 
```

What is newer is [namespace map syntax introduced in Clojure 1.9](https://clojure.atlassian.net/browse/CLJ-1910).


```Clojure
❯ clj -Sdeps '{:deps {org.clojure/clojure {:mvn/version "1.9.0"}}}' 
Clojure 1.9.0
user=> #::{:foo 1}
#:user{:foo 1}
user=> 
```

And Clojure spec is 1.9+ too.

[...]

PR #194, opened 2023-01-23
Adapt library to support Clojure 1.8

philomates:
Since it is easy enough to do
Related to https://github.com/nubank/matcher-combinators/issues/193


philomates:
@lread would you be up to review this change as you're familiar with these setups in other projects?

deps.edn
```diff
@@ -2,15 +2,18 @@
              "clojars"  {:url "https://clojars.org/repo/"}}
  :paths ["src/clj" "src/cljc" "src/cljs"]
  :deps {org.clojure/clojure {:mvn/version "1.11.1"}
```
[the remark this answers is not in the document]
lread: Yeah, I think so @philomates, if this ends up in the pom.
If a library pom specifies Clojure as a dep, I think it should specify the min required version.

.github/workflows/clojure.yml
```diff
@@ -40,11 +40,23 @@ jobs:
     - name: Run midje tests
       run: bb test:midje
 
-    - name: Run clj tests
-      run: bb test:clj
+    - name: Run clj tests with Clojure 1.8
+      run: bb test:clj -A:1.8
```
lread, line 45: So in bb.edn we have:

`(apply clojure "-A:dev:clj-test:test-runner" *command-line-args*)`

Which means we can specify `-A` multiple times? I did not know/remember that.

A bit pedantic maybe but: `-M` has replaced `-A`. So probably should be using `-M` instead of `-A`.

philomates: hmmm, yeah, I see those CLI warnings about `-A` being deprecated.
I tried replacing `-A` with `-M` in `clj:test`'s `(apply clojure "-A:dev:clj-test:test-runner" *command-line-args*)` and it makes it so passing an additional `-A:1.9` doesn't work :/

lread: So if my wee bit of experiment at the terminal is right, it seems multiple `-A`s concatenate whereas for multiple `-M`s the first one wins. 

Your repeating `-A` trick seems to keep the code nice and simple at the cost of a warning and (perhaps) taking advantage of an obscure behaviour.

Much less verbose than the little wrappers I tend to write. [Here's one as an example if you are interested](https://github.com/clj-commons/clj-yaml/blob/4bd5bdf1c068cec31ba96e3089f34593f59b7875/script/test_clj.clj).

philomates:
Thanks for the review @lread! I incorporated all the suggestions except the `-A`/`-M` change; figured it is okay debt for now that can be addressed if the build becomes more complex

Issue #202, opened 2023-03-07
`matchers/match-with` misbehaves for CLJS numbers


bryanluu:
When playing around with the `matchers/match-with` method in a CLJS test to set numeric matchers to use `matchers/within-delta`, I discovered an interesting bug. When using `matchers/within-delta` directly, the test behaves as expected. However, when using `matchers/match-with [number? (matchers/within-delta delta)] expected`, a clojurescript test would fail with the message: 

> Error: No protocol method Matcher.-matcher-for defined for type number ,,,

The following test illustrates this issue.
```clojure
(ns example-test
  (:require [matcher-combinators.test] ; adds support for `match?` and `thrown-match?` in `is` expressions
            [matcher-combinators.matchers :as m]
            #?(:clj [clojure.test :refer [deftest testing is]]
               :cljs [cljs.test :refer-macros [deftest testing is]])))

(deftest failing-matcher-test
  (testing "Example numeric test-case"
    (testing "which passes"
      (is (match? (m/within-delta 0.05 1) 0.99)))
    (testing "which errors"
      (is (match? (m/match-with [number? (m/within-delta 0.05)] 1) 0.99)))))
```

I am using `nubank/matcher-combinators {:mvn/version "3.4.0"}`. This behaviour seems to only affect the `number?` checks.

philomates:
Hey, thanks for the issue + reproduction steps.
I found the main issue and am patching it in https://github.com/nubank/matcher-combinators/pull/203 (still a work in progress)

philomates:
I released a fix in 3.8.6, sorry for the delay

PR #203, opened 2023-03-13
WIP: fix match-with for cljs

philomates:
addresses https://github.com/nubank/matcher-combinators/issues/202

it uncovered an issue with `pred` when used in `match-with` that I still need to figure out (tests are currently failing because of this)


NoahTheDuke:
Looks like the tests aren't returning a non-0 status so CI isn't failing but the tests are failing:

https://github.com/nubank/matcher-combinators/actions/runs/5586358717/job/15128852733

PR #204, opened 2023-03-24
Configure clj-kondo for ClojureScript as well

ferdinand-beyer:
The exported clj-kondo configuration currently only considers Clojure's `clojure.test`, but not ClojureScript's `cljs.test`. Therefore you'll get linter errors when using `match?` in ClojureScript tests.


philomates (approved):
nice, thanks for the patch

philomates:
released as `3.8.5`

PR #207, opened 2023-05-03
Change stubs to throw exceptions if used incorrectly

### LINKED ISSUE #206: Bad usage stubs for `match?` and `thrown-match?`
`clojure.test/assert-expr` is clever but slightly magical, which makes it harder to work with: Where does `match?` come from? How does it work? Why aren't there any docstrings? Why does Clojure throw an "unable to resolve symbol: match?" error if I use this outside of `is`?

My proposed solution (as seen in [clojure-expectations.clojure-test](https://github.com/clojure-expectations/clojure-test)) is to create dummy functions that can be referred but will throw exceptions if used outside of `is` calls:

```clojure
(defmacro bad-usage [msg]
  `(throw IllegalArgumentException ~msg))

(defn match?
  "The `matcher` should be the matcher-combinator represented the expected value,
   and the `actual` should be the expression being checked."
  [matcher actual]
  (bad-usage "match? must be used inside of `is`.))
```

This way, a user can write:

```clojure
(ns example.foo
  (:require
    [clojure.test :refer [deftest is]]
    [matcher-combinators.test :refer [match?]]
    [matcher-combinators.matchers :as m]))

(deftest test-matching-with-explicit-matchers
  (match? (m/equals 37) (+ 29 8))
  (is (match? (m/regex #"fox") "The quick brown fox jumps over the lazy dog")))
```

And they'll get a reasonable error message: ``"match? must be used inside of `is`."`` instead of the less helpful message `Syntax error compiling at (example/foo.clj). Unable to resolve symbol: match? in this context.`

If this is reasonable, I'd be willing to open a PR with these changes.

NoahTheDuke:
Closes #206

Implements the `bad-usage` macro as described in #206.

I've opened this as a draft because for some reason, I can't get the clojurescript half of this to work. Running the tests with `bb test:node` shows these errors, making me doubt that the tests are running correctly:

```
[:cli-test] Build completed. (76 files, 3 compiled, 6 warnings, 4.25s)

------ WARNING #1 - :undeclared-var --------------------------------------------
 File: /home/noah/personal/matcher-combinators/test/cljs/matcher_combinators/cljs_example_test.cljs:60:16
--------------------------------------------------------------------------------
  57 | 
  58 | (deftest bad-usage-test
  59 |   (is (thrown? js/Error
  60 |                (match? :expected :actual)))
----------------------^---------------------------------------------------------
 Use of undeclared Var matcher-combinators.cljs-example-test/java
--------------------------------------------------------------------------------
...
```

I don't see see how what I wrote is any different than what's in clojure-expectations, so I'm pretty confused.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ This library addresses this issue by providing composable matcher combinators th
 
 ### `clojure.test`
 
-Require the `matcher-combinators.test` namespace, which will extend `clojure.test`'s `is` macro to accept the `match?` and `thrown-match?` directives.
+Refer `match?` and `thrown-match?` from the `matcher-combinators.test`:
 
  - `match?`: The first argument should be the matcher-combinator represented the expected value, and the second argument should be the expression being checked.
  - `thrown-match?`: The first argument should be a throwable subclass, the second a matcher-combinators, and the third the expression being checked.
@@ -46,7 +46,7 @@ For example:
 
 ```clojure
 (require '[clojure.test :refer [deftest is]]
-         '[matcher-combinators.test] ;; adds support for `match?` and `thrown-match?` in `is` expressions
+         '[matcher-combinators.test :refer [match? thrown-match?]]
          '[matcher-combinators.matchers :as m])
 
 (deftest test-matching-with-explicit-matchers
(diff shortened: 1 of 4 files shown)


NoahTheDuke:
I've poked at this some more and I have no idea why there's an "undeclared-var" warning happening.

PR #208, opened 2023-05-04
Extra info for seq-of

kdmoreira:
Matcher `seq-of` expects a non-empty sequence, but it wasn't clear from the docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,7 +176,7 @@ for a specific value, e.g.
 
 - `set-equals`/`set-embeds` similar behavior to `equals`/`embeds` for sets, but allows one to specify the matchers using a sequence so that duplicate matchers are not removed. For example, `(equals #{odd? odd?})` becomes `(equals #{odd})`, so to get around this one should use `(set-equals [odd? odd])`.
 
-- `seq-of` takes an expected matcher and creates a new matcher over a sequence, where each element matches the provided expected matcher. Analogous to `clojure.core/every?`.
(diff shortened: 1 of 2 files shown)


dchelimsky:
Thinking this through a bit more since first bringing it up: another possibility is to let it validate an empty sequence. That would align with Clojure's handling of sequences in general e.g. `(map f []) ;; => ()`

An argument against doing this is that it could lead to false positives.

WDYT?

dchelimsky:
> Thinking this through a bit more since first bringing it up: another possibility is to let it validate an empty sequence. That would align with Clojure's handling of sequences in general e.g. `(map f []) ;; => ()`
> 
> An argument against doing this is that it could lead to false positives.
> 
> WDYT?

Also:

```clj
> (every? odd? [])
true
```

philomates:
I'd be up for doing a breaking change (4.0.0?) to fix this to be in-line with clojure's behavior

dchelimsky:
We could also deprecate `seq-of` and add an `every` matcher to avoid a breaking change.

philomates:
I kind of liked that we had these `*-of` matchers: `any-of`, `all-of`, and `seq-of`.
If we rename `seq-of` to `every`, should we also rename `any-of` and `all-of`? If so do you have good name suggestions that consider https://github.com/nubank/matcher-combinators/pull/191#discussion_r1090881179 ?
There is also the downside of shadowing `clojure.core` with `every`.

Issue #159, opened 2021-06-01
Option for not printing full `expected` and `actual` in mismatch message


philomates:
Currently when a mismatch happens (in clojure.test) we use the entire `(match? the-expected the-actual)` form as the `Expected: ` field and the annotated mismatch diff as the `Actual: ` field. This means that if `the-expected` or `the-actual` matcher/data-structures are large, we get really big output.

In some cases, like when you are matching and expected file contents with an actual file contents, it really suffices to only see the precise mismatch info. This is partially the motivation behind https://github.com/nubank/matcher-combinators/pull/157.

What would be useful to look into is a way to abbreviate or omit aspects of the output.

philomates:
one suggestion by @gusbicalho is an annotation construct:

```clojure
(match?
  (-> expected (reported-as "contents of foo.graphql"))
  (-> actual (reported-as "contents of tmp-file.graphql")))
```

philomates:
another option would be to limit the size of the `Expected: ` output at print time. Something like https://github.com/nubank/matcher-combinators/compare/expected-printing?expand=1# which would for
```clojure
(is (match? [1 (m/not-matcher odd?) 3]
            [1 2 3]))
```
output
```
FAIL in (absent-pred) (test_test.clj:120)
expected: (match? [1 ...)
  actual: [1
 (mismatch
  "expected mismatch from: clojure.core$even_QMARK_@6a0a7d89"
  2)
 3]
```

gabrielgiussi:
Couldn't we just traverse the result and redact/remove anything that is not a Mismatch?
If we run this
```clojure
(is (match? {:a {:b odd?}}
              {:a {:b 2
                   :c 3}}))
```

We have this result
```clojure
{:matcher-combinators.result/type :mismatch,
 :matcher-combinators.result/value {:a {:b #matcher_combinators.model.Mismatch{:expected (pred
                                                                                          #object[matcher_combinators.matchers$pred
                                                                                                  0x46414468
                                                                                                  "matcher_combinators.matchers$pred@46414468"]),
                                                                               :actual 2},
                                        :c 3}},
 :matcher-combinators.result/weight 1}
```

I would like to see
```clojure
  actual: {:a
 {:b
  (mismatch
   (expected
    (pred
          #object[matcher_combinators.matchers$pred 0x46414468 "matcher_combinators.matchers$pred@46414468"]))
   (actual 2))}}
; c was removed completely
```

For other collections where the order is important we could have something like 
```clojure
[...]

PR #214, opened 2023-08-31
Redact fully-matched data-structures from output

philomates:
addresses https://github.com/nubank/matcher-combinators/issues/159 by allowing user to manually turn on output redaction, which removes all fully-matched data-structures from the output. When output is redacted a `'...` entry is added to the top level mismatching data-structure. This helps keep the context that what you are seeing isn't the actual full form.

### For maps
```clojure
(matcher-combinators.config/disable-abbreviation!)
(is (match? {:a [1 2 {:a 2 :b [4 5] :c [2 [3 4]]}]}
            {:a [1 2 {:a 2 :b [4 6] :c [2 [3 4]]}]})
    "this is standard output")

(matcher-combinators.config/enable-abbreviation!)
(is (match? {:a [1 2 {:a 2 :b [4 5] :c [2 [3 4]]}]}
            {:a [1 2 {:a 2 :b [4 6] :c [2 [3 4]]}]})
    "this is abbreviated")
```
we get:
20230901_13h25m55s_grim

### for sequences

```clojure
(matcher-combinators.config/disable-abbreviation!)
(is (match? [1 2 {:a 2 :b [4 5] :c [2 [3 4]]}]
            [1 2 {:a 2 :b [4 6] :c [2 [3 4]]}])
    "this is standard output")

(matcher-combinators.config/enable-abbreviation!)
(is (match? [1 2 {:a 2 :b [4 5] :c [2 [3 4]]}]
            [1 2 {:a 2 :b [4 6] :c [2 [3 4]]}])
    "this is abbreviated output")
```

20230901_13h25m34s_grim

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@ All notable changes to this project will be documented in this file. This
 change log follows the conventions of
 [keepachangelog.com](http://keepachangelog.com/).
 
+## 3.8.7 / 2023-08-31
+- [Experimental] add experimental feature to not print the entire actual data-structure on
+  mismatch, but only the mismatches.
+
 ## 3.8.6 / 2023-07-18
 - fix issue when using non-composite matchers (`m/regex`, `m/pred`, etc)
   inside `match-with`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,11 @@ change log follows the conventions of
 [keepachangelog.com](http://keepachangelog.com/).
 
 ## 3.8.7 / 2023-09-01
-- [Experimental] add experimental feature to not print the entire actual data-structure on
-  mismatch, but only the mismatches.
+- introduce `matcher-combinators.config` namespace to toggle use of ansi color
+  codes and the new output abbreviation mode.
+- [Experimental] add `(matcher-combinators.config/enable-abbreviation!)`, an
+  experimental feature to print only the mismatched parts of a data-structure
+  while elliding the matched parts.
 - fix more issues when using non-composite matchers (`m/regex`, `m/pred`, etc)
   inside `match-with`.
 
--- a/src/cljc/matcher_combinators/ansi_color.cljc
+++ b/src/cljc/matcher_combinators/ansi_color.cljc
@@ -7,19 +7,20 @@
   *use-color*
   true)
 
-(defn- set-use-color! [v]
-  #?(:clj (alter-var-root #'*use-color* (constantly v))
-     :cljs (set! *use-color* v)))
-
-(defn enable!
-  "Thread-global way to enable the usage of ANSI color codes in matcher-combinator output."
+(defn
+  ^{:deprecated true
+    :doc "DEPRECATED! Use matcher-combinators.config/enable-ansi-color!"}
+  enable!
   []
-  (set-use-color! true))
+  #?(:clj (alter-var-root #'*use-color* (constantly true))
+     :cljs (set! *use-color* true)))
 
 (defn disable!
-  "Thread-global way to disable the usage of ANSI color codes in matcher-combinator output."
+  ^{:deprecated true
+    :doc "DEPRECATED! Use matcher-combinators.config/disable-ansi-color!"}
   []
-  (set-use-color! false))
+  #?(:clj (alter-var-root #'*use-color* (constantly false))
+     :cljs (set! *use-color* false)))
 
 (def ANSI-CODES
   {:reset              "[0m"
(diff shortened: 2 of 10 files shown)


src/cljc/matcher_combinators/ansi_color.cljc
```diff
@@ -7,19 +7,20 @@
   *use-color*
```
philomates, line 7: would be nice to move this also to `matcher-combinators.config` but would be breaking in case anybody uses it with `binding`

src/cljc/matcher_combinators/config.cljc
```diff
@@ -0,0 +1,47 @@
+(ns matcher-combinators.config
+  "Global output behavior configurations"
```
philomates, line 2: figure at some point we could also implement something that reads `~/.config/clojure/matcher-combinators.edn` config file. But for now folks can include `(matcher-combinators/enable-redaction!)` in their `dev/user.clj` startup file to get the same behavor

CHANGELOG.md
```diff
@@ -3,6 +3,10 @@ All notable changes to this project will be documented in this file. This
 change log follows the conventions of
 [keepachangelog.com](http://keepachangelog.com/).
 
+## 3.8.7 / 2023-08-31
+- [Experimental] add experimental feature to not print the entire actual data-structure on
```
philomates, line 7: marking it as experimental because I haven't really tested it much and would be good to get some user feedback before committing to the feature. Sound alright or someone have another idea?

acamargo:
I'm in very need of this feature :-) Would be possible to store the full payload on a temp file and add the path in the error message? This way we would be communicating that it is being used the redacted version and it allows us to inspect the file in case we really need to take a look into it.

philomates:
@NoahTheDuke glad to hear you'd find it useful. And I managed to add `...`s to the redacted output.
@acamargo interesting; we could also bind the full data to a var so you could look it up after the fact, similar to how the repl binds results to `*1`, `*2`, etc. For now though I'd like to keep it simple and see if the need really arises (you can always re-run the test with redaction turned off).

philomates:
decided to go with "abbreviated" instead of "redacted" when referring to the feature https://github.com/nubank/matcher-combinators/pull/214/commits/7ef0b2d2b7592053c386ab5d78ece6704e617bc9

philomates:
@gabrielgiussi when you have a chance would be nice to get a review from you

gabrielgiussi:
I think this impl is great, simple but giving us enough to make mismatches so much easier to read with large datastructures. Great work!

I only found slightly confusing the double `... ...` in the output and I played a bit around with a way to avoid it.
It can be done by introducing a new record `EmptyMarker` and do this assoc instead `(assoc expr ellision-marker empty-marker)`.
Besides that we need to skip it in `print-diff-dispatch`, so maybe doesn't worth the effort.

gabrielgiussi (approved):

philomates:
I also found the `{... ...}` a bit weird in the map; replaced it with an empty marker, like you suggested

gabrielgiussi:
Another thing I forget to mention...
The whole data structure is still in the expected value, which can still flood the console output.
I think we could also check the `*use-abbreviation*` var [here](https://github.com/nubank/matcher-combinators/blob/3df9b17e8721b6d5542ceda5f4de2a3e3be1fc43/src/clj/matcher_combinators/clj_test.clj#L61C3-L61C3) and simply replace the whole expected with a message.
WDYT?

philomates:
> Another thing I forget to mention... The whole data structure is still in the expected value, which can still flood the console output. I think we could also check the `*use-abbreviation*` var [here](https://github.com/nubank/matcher-combinators/blob/3df9b17e8721b6d5542ceda5f4de2a3e3be1fc43/src/clj/matcher_combinators/clj_test.clj#L61C3-L61C3) and simply replace the whole expected with a message. WDYT?

@gabrielgiussi can you provide an example of still seeing the `expected` value?
I'm not seeing it w/ my setup. Like I see something like:

```
FAIL in (test-compound) (workflow_test.clj:450)
expected: (match? (m/in-any-order [{:service/code "wash"} {:service/code "wax on"} {:service/code "chop wood"} {:service/code "carry water"}]) 
                  (some-expression))
  actual: (... (unexpected {:service/code "wax off", :service/fahrzeug 'something}))
```

Do you mean that we see `(some-expression)`?

gabrielgiussi:
I just did a simple example the other day to announce in our slack about this new feature,
```clojure
(deftest abbreviate-test
  (is (match? {:a 1}
              {:a 2 :b :huge-map})))
```

But now you show an example with the fn call I just realize that's how it is actually used so showing that as the expected is just fine. I forgot we are showing the form in the expected.

Issue #217, opened 2024-01-24
Common misconception about `equals` not being applied to `map`s


gilvan-reis:
# Context
It is a common misconception from users that the default matcher for `map` is `embeds`. We already had another [issue](https://github.com/nubank/matcher-combinators/issues/108) that improved the documentation, but this problem continues to happen very often.

In my team, we are creating several replicated functions to do `deep-equals` (following this README [section](https://github.com/nubank/matcher-combinators?tab=readme-ov-file#overriding-default-matchers)) to validate nested `map`s.

This discussion already started in a [message](https://clojurians.slack.com/archives/C04ABMF89D3/p1704995012874339) in the #matcher-combinator slack channel.

# Proposed Solution
Create the `deep-equals` function in the `matcher-combinators.matchers` namespace with a proper docstring trying to bring this explanation closer to the code.

```clojure
(defn deep-equals [expected]
  (match-with [map? m/equals]
              expected))
```

gilvan-reis:
Closed by https://github.com/nubank/matcher-combinators/pull/218

PR #218, opened 2024-01-24
Create nested-equals matcher

gilvan-reis:
Create `nested-equals` function to match nested maps with `equals` and providing a documentation closer to the code about the nested behavior for matchers.

See https://github.com/nubank/matcher-combinators/issues/217

### DIFF
--- a/src/cljc/matcher_combinators/matchers.cljc
+++ b/src/cljc/matcher_combinators/matchers.cljc
@@ -34,13 +34,17 @@
 (declare match-with)
 
 (defn strictly-equals
-  "Return a matcher for the expected value which matches nested maps with `equals`.
-  This solves a common need of matching nested maps with strict equality.
-  See also: `match-with`.
+  "A matcher that always uses the `equals` matcher at every level of nesting.
 
-  Usage:
-    (is (match? (strictly-equals {:a :b})
-                {:a :b :c :d}))"
+  Useful given that matchers usually only change the first level of the data
+  they are applied to, leaving nested data to use the default matcher of that
+  type of data. For instance, this can be used to assert that any nested map
+  has exactly the same keys and matching values as provided in the `expected`,
+  and no more.
+
+  Note: this excludes functions, which continue to be invoked
+  as predicates, and regex, which continue to be invoked with `regex`,
+  instead of being compared via the `equals` matcher."
   [expected]
   (match-with [map? equals] expected))
 
--- a/src/cljc/matcher_combinators/matchers.cljc
+++ b/src/cljc/matcher_combinators/matchers.cljc
@@ -31,6 +31,19 @@
         :else
         (core/->Value expected)))
 
+(declare match-with)
+
+(defn deep-equals
+  "Return a matcher for the expected value which matches nested maps with `equals`.
+  This solves a common need of matching nested maps with strict equality.
+  See also: `match-with`.
+
+  Usage:
+    (is (match? (deep-equals {:a :b})
+                {:a :b :c :d}))"
+  [expected]
+  (match-with [map? equals] expected))
+
 (defn seq-of
   "Matcher that will match when given a sequence where every element matches
   the provided `expected` matcher. It expects a non-empty sequence."
@@ -183,8 +196,6 @@
            first
            last)
       (matcher-for value)))
-
-(declare match-with)
 
 (defn- match-with-values [m overrides]
   (reduce-kv (fn [m* k v] (assoc m* k (match-with overrides v)))
--- a/src/cljc/matcher_combinators/matchers.cljc
+++ b/src/cljc/matcher_combinators/matchers.cljc
@@ -33,7 +33,7 @@
 
 (declare match-with)
 
-(defn strictly-equals
+(defn nested-equals
   "A matcher that always uses the `equals` matcher at every level of nesting.
 
   Useful given that matchers usually only change the first level of the data
--- a/src/cljc/matcher_combinators/matchers.cljc
+++ b/src/cljc/matcher_combinators/matchers.cljc
@@ -33,13 +33,13 @@
 
 (declare match-with)
 
-(defn deep-equals
+(defn strictly-equals
   "Return a matcher for the expected value which matches nested maps with `equals`.
   This solves a common need of matching nested maps with strict equality.
   See also: `match-with`.
 
   Usage:
-    (is (match? (deep-equals {:a :b})
+    (is (match? (strictly-equals {:a :b})
                 {:a :b :c :d}))"
   [expected]
   (match-with [map? equals] expected))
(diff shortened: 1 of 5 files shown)


src/cljc/matcher_combinators/matchers.cljc
```diff
@@ -31,6 +31,19 @@
         :else
         (core/->Value expected)))
 
+(declare match-with)
+
+(defn deep-equals
+  "Return a matcher for the expected value which matches nested maps with `equals`.
+  This solves a common need of matching nested maps with strict equality.
+  See also: `match-with`.
```
gilvan-reis, line 39: Feel free to suggest a better explanation here

src/cljc/matcher_combinators/matchers.cljc
```diff
@@ -31,6 +31,19 @@
         :else
         (core/->Value expected)))
 
+(declare match-with)
+
+(defn deep-equals
```
gilvan-reis, line 36: ❓  question (non_blocking): Do you think we should create a test for it? It would be very similar to the `match-with` [test](https://github.com/nubank/matcher-combinators/blob/ccd20994884733860b7995f418b065bfee3c3c55/test/clj/matcher_combinators/matchers_test.clj#L260)

scottbale: I would be in favor of a test. Even if it is a little redundant, if nothing else it serves as additional documentation.

philomates:
other possible names for `deep-equals`:
- `equals-as-default`: has the benefit where `default` keys the reader to the fact that there is a default that isn't always `equals`.
- `always-equals`
- `strict`

As a note, the midje implementation had a similar thing, [`match-equals`](https://github.com/nubank/matcher-combinators/blob/a5f00f60ce575266a519ddbcc090d7a2e749a3f6/src/clj/matcher_combinators/midje.clj#L144-L148), but the name isn't so great because it conflates the `match` construct with the combinators.

gilvan-reis:
I think we should have `equals` in the function name, so I decided to follow with `strictly-equals` WDYT?

src/cljc/matcher_combinators/matchers.cljc
```diff
@@ -31,6 +31,19 @@
...
+(declare match-with)
+
+(defn strictly-equals
+  "Return a matcher for the expected value which matches nested maps with `equals`.
+  This solves a common need of matching nested maps with strict equality.
+  See also: `match-with`.
+
+  Usage:
+    (is (match? (strictly-equals {:a :b})
+                {:a :b :c :d}))"
```
philomates, line 43: this example isn't a case where using `strictly-equals` would be any different than from using `equals` at the top level.
I'd go with something that shows the difference from

```clojure
(is (match? (m/equals {:user1 {:id 5}
                       :user2 {:id 3}})
            {:user1 {:id 5 :name "hennix"}
             :user2 {:id 3 :name "flynt"}}))
```
which gives different results than
```clojure
(is (match? (m/equals {:user1 (m/equals {:id 5})
                       :user2 (m/equals {:id 3})})
            {:user1 {:id 5 :name "hennix"}
             :user2 {:id 3 :name "flynt"}}))
```

src/cljc/matcher_combinators/matchers.cljc
philomates: Maybe something more like the following which explains it in broader terms and mentions the fact that there are defaults that are being overridden:
> A matcher that always uses the `equals` matcher at every level of nesting. Useful given that matchers usually only change the first level of the data they are applied to, leaving nested data to use the default matcher of that type of data. For instance, this can be used to assert that any nested map has exactly the same keys and matching values as provided in the `expected`, and no more.
>
> Note: this excludes functions, which continue to be invoked as predicates instead of compared via the `equals` matcher

test/clj/matcher_combinators/matchers_test.clj
```diff
@@ -361,6 +361,23 @@
...
+                {:a :b :c {:d {:e :f}}})))
+
+  (testing "failing case with one level map"
+    (is (no-match? (m/strictly-equals {:a :b})
+                   {:a :b :c :d})))
+
+  (testing "failing case with multi level map"
+    (is (no-match? (m/strictly-equals {:a :b :c {:d {:e :f}}})
+                   {:a :b :c {:d :e}}))))
+
```
philomates, line 380: can you add a case to show how functions are treated (like still as predicates and not as `equals` matchers)
something like
```clojure
(is (match? (m/strictly-equals {:x odd?})
            {:x 1}))
```

test/clj/matcher_combinators/matchers_test.clj
```diff
@@ -361,6 +361,23 @@
...
+    (is (match? (m/strictly-equals {:a :b :c {:d {:e :f}}})
+                {:a :b :c {:d {:e :f}}})))
+
+  (testing "failing case with one level map"
+    (is (no-match? (m/strictly-equals {:a :b})
+                   {:a :b :c :d})))
+
+  (testing "failing case with multi level map"
+    (is (no-match? (m/strictly-equals {:a :b :c {:d {:e :f}}})
+                   {:a :b :c {:d :e}}))))
```
philomates, line 379: this would also fail without the `strictly-equals`
Maybe something like 
```clojure
(is (match? (m/strictly-equals {:a :b :c {:d :e}})
            {:a :b :c {:d :e
                       :y :z}}))
```

philomates (commented):
Nice, thanks for the PR!

Can you add the matcher to the list in the readme?: https://github.com/nubank/matcher-combinators#built-in-matchers

Also the https://github.com/nubank/matcher-combinators#overriding-default-matchers mentions a `match-with` example that could be replaced with `strictly-equals`. Could you add a line there that says "For convenience we've also added this as a built-in matcher called `strictly-equals`?

src/cljc/matcher_combinators/matchers.cljc
gilvan-reis: Yeah, I was thinking to put this case just to inform it maintain the `equals` behavior (similar to using the `match-with`), but I will change it to show the cases where it would be helpful

src/cljc/matcher_combinators/matchers.cljc
gilvan-reis: I have added `regex` exclusion too. Feel free to correct me if I'm wrong

gilvan-reis:
@philomates I have replaced the `match-with` example in the README. Feel free to inform me if you want to stay with both instead

gilvan-reis:
I have also added the versions changes to be validated

scottbale (approved):
LGTM but I'll defer to @philomates's review

philomates (approved):
Looks good!
Was mulling the name over a little bit more and `nested-equals` came up. I like it because it refers to nesting, which is really where this differs from the default. Like the matcher used for functions and regex stays the same, but for any nested/container datastructure equals is used instead of whatever default.

What do you think? Or should we just stick with this?

PR #220, opened 2024-02-08
Bump clojure tools versions

gilvan-reis:
Even after [adding the license](https://github.com/nubank/matcher-combinators/pull/219) to the POM file, we still [can not deploy to clojars](https://github.com/nubank/matcher-combinators/actions/runs/7820320633/job/21334888472).

After some analysis, I discovered the problem was related to the `tools.deps` version (`0.16.1264`) being used that does not support the usage of the `:pom-data`. 

So I'm bumping the `tools.build` version to generate the correct POM file. I could verify that everything was correct by executing `bb jar` and checking the file on `target/classes/META-INF/maven/nubank/matcher-combinators/pom.xml`

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -36,6 +36,6 @@
 
   :build
   {:extra-deps {slipset/deps-deploy {:mvn/version "0.2.0"}
-                io.github.clojure/tools.build {:git/sha "fe6b1405ba888720c813c7488c03880be73bbe20"}}
+                io.github.clojure/tools.build {:mvn/version "0.9.6"}}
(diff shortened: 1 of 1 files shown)


scottbale (approved):
I can't think of any problems with bumping this dep, LGTM.

philomates:
so it required a version downgrade of `tools.deps`? If the build runs, then :shipit: 
thanks for investigating

gilvan-reis:
@philomates Actually it was a upgrade of `tools.deps` from `0.16.1264` to `0.18.1354`

Issue #222, opened 2024-04-29
Case when enabling abbreviation results in no useful mismatch info


philomates:
The abbreviation feature (added in https://github.com/nubank/matcher-combinators/pull/214) seems to show no useful output for the following example

```clojure
(require '[matcher-combinators.test :refer [match?]])
(require '[clojure.test :refer [is]])

;; normal behavior
(matcher-combinators.config/disable-abbreviation!)
(is (match? [#{2}] [#{1 2}]))
;; =>
;; expected: (match? [#{2}] [#{1 2}])
;;   actual: [#{(unexpected 1) 2}]

;; current abbreviation behavior
(matcher-combinators.config/enable-abbreviation!)
(is (match? [#{2}] [#{1 2}]))
;; =>
;; expected: (match? [#{2}] [#{1 2}])
;;   actual: [...]
```

I would expect the abbreviation feature, when enabled, to give something more like for the `(is (match? [#{2}] [#{1 2}]))` mismatch:
```clojure
;; expected: (match? [#{2}] [#{1 2}])
;;   actual: [#{(unexpected 1) ...}]
```

PR #225, opened 2025-01-17
Experimental ellision feature fix: Preserve metadata on sets to make it work with feature

philomates:
fixes https://github.com/nubank/matcher-combinators/issues/222

### DIFF
--- a/src/cljc/matcher_combinators/core.cljc
+++ b/src/cljc/matcher_combinators/core.cljc
@@ -246,7 +246,8 @@
 
 (defn- type-preserving-mismatch [base-list values]
   (let [lst (into base-list values)]
-    (if (vector? base-list)
+    (if (or (vector? base-list)
+            (set? base-list))
       lst
       (reverse lst))))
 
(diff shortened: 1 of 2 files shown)


src/cljc/matcher_combinators/core.cljc
```diff
@@ -246,7 +246,8 @@
 
 (defn- type-preserving-mismatch [base-list values]
   (let [lst (into base-list values)]
-    (if (vector? base-list)
+    (if (or (vector? base-list)
+            (set? base-list))
       lst
       (reverse lst))))
```
philomates, line 252: without this change we were turning a set into a list due to `reverse`

src/cljc/matcher_combinators/core.cljc
```diff
@@ -397,9 +398,11 @@
...
        ::result/value  elements
        ::result/weight 0}
-      (match (->EqualsSeq (concat (:matched result)
-                                  (:unmatched result)))
-             (:elements result)))))
+      (update (match (->EqualsSeq (concat (:matched result)
+                                          (:unmatched result)))
+                     (:elements result))
+              ::result/value
+              #(with-mismatch-meta % :mismatch-sequence)))))
```
philomates, line 405: add `:mismatch-sequence` metadata, which is used by the elision feature, to the result of any-order/set mismatch results

PR #236, opened 2026-04-13
Add docstrings to matcher-combinators.test declarations

philomates:
I had to change the `declare` forms in `matcher-combinators.test` to `defn`s so that the docstring would be picked up by LSP. Just adding `:doc` meta to the declare, like the following, doesn't seem to get picked up by the clojure LSP as far as I can tell
```clojure
(require '[matcher-combinators.test :refer [match? thrown-match?]])
(declare ^{:arglists '([matcher actual]) :doc "some docstring"} match?)
(:doc (meta #'match?))
```

### DIFF
--- a/src/cljc/matcher_combinators/test.cljc
+++ b/src/cljc/matcher_combinators/test.cljc
@@ -15,19 +15,43 @@
   Commonly, a dev-only user namespace will require this namespace."
   (:require
    #?(:cljs [cljs.test    :as t :refer-macros [is are deftest testing]]
-       :clj  [clojure.test :as t :refer        [is are deftest testing]])
-    #?(:cljs [matcher-combinators.cljs-test]
-       :clj  [matcher-combinators.clj-test])))
+      :clj  [clojure.test :as t :refer        [is are deftest testing]])
+   #?(:cljs [matcher-combinators.cljs-test]
+      :clj  [matcher-combinators.clj-test])))
 
-(declare ^{:arglists '([matcher actual])}
-         match?)
-(declare ^{:arglists '([type->matcher matcher actual])}
-         match-with?)
-(declare ^{:arglists '([matcher actual]
-                       [exception-class matcher actual])}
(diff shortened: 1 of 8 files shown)


src/cljc/matcher_combinators/test.cljc
```diff
@@ -15,19 +15,43 @@
...
+(is (match? (matcher-combinators.matchers/in-any-order [zero? odd? even?]) (range 3)))"
+  [matcher actual]
+  (throw (#?(:cljs js/Error. :clj AssertionError.)
+          "Should only be invoked within a `clojure.test/is` form")))
+
+(defn match-with?
+  "DEPRECATED: Use (match? (matcher-combinators.matchers/match-with <type->matcher> <expected>) <actual>) instead."
+  [type->matcher matcher actual]
+  (throw (#?(:cljs js/Error. :clj AssertionError.)
+          "Should only be invoked within a `clojure.test/is` form")))
```
teodorlu, line 37: Could also be helpful for users to include `:deprecated "deprecated-version"` metadata:

```suggestion
(defn match-with?
  "DEPRECATED: Use (match? (matcher-combinators.matchers/match-with <type->matcher> <expected>) <actual>) instead."
  {:deprecated "3.0.0"}
  [type->matcher matcher actual]
  (throw (#?(:cljs js/Error. :clj AssertionError.)
          "Should only be invoked within a `clojure.test/is` form")))
```

This will show up in most editors that rely on `clojure-lsp`. My Emacs the referenced var with strikethough, and deprecated version below:

Image

NoahTheDuke:
Looks like this is a redo of my PR #207, which i couldn't ever get to work. I'll close that one if you can get the test suite to pass.

teodorlu:
Thank you! 🙏

philomates:
released as `3.10.1`

NoahTheDuke:
thank you!

PR #237, opened 2026-06-29
Add sorted order

marcojunior-code:
Adds a new sorted-by matcher that compares two sequences regardless of order by sorting both expected and actual with a caller-supplied key-fn and then matching them positionally (like equals over sequences).                                                                                                          

It's an O(n log n) alternative to in-any-order, which searches for a valid pairing between matchers and elements and can be prohibitively slow on large sequences. When you have a cheap, stable sort key, sorted-by avoids that search entirely.                                                                         

Motivation                                                                                                                                                                                                                                                                                                                
in-any-order is the existing tool for order-insensitive matching, but its pairing search is expensive for large collections. For the common case of sequences of concrete, uniformly-typed, sortable values, sorting both sides by a key is far cheaper and just as correct.                                              
                                                                                                        
What changed                                                                                                                                                                                                                                                                                                              

- core.cljc — new SortedBy record implementing the Matcher protocol. It validates that actual is sequential?, then sorts expected and actual by key-fn and delegates to sequence-match (strict, equals-like — extra/missing elements mismatch).                                                                           
- matchers.cljc — public sorted-by constructor with thorough docstring covering the precondition and limitation (below).                                                                                                                                                                                                  
- README.md — documents sorted-by alongside in-any-order.                                                                                                                                                                                                                                                                 
- matchers_test.clj — tests covering concrete-value matching, submatchers with unique keys, mismatch detection, missing-element detection, invalid-type handling, match-with integration, and explicit tests documenting the precondition/limitation edge cases.                                                          

Caveats (documented in the docstring, README, and tests)                                                                                                                                                                                                                                                                  

- Precondition — key-fn must induce a total order. It's the caller's responsibility (analogous to clojure.core/sort needing a consistent comparator):                                                                                                                                                                     
- Tied keys leave equal-keyed elements in incoming order, which can differ between the two sides and cause a spurious mismatch — fix with a compound key, e.g. (juxt :x :id).                                                                                                                                           
- Mixed-type keys make sort-by throw (e.g. ClassCastException), by design — it signals misuse rather than a mismatch.                                                                                                                                                                                                   
- Limitation — not suitable with submatchers in general. When expected contains submatchers (predicates, equals, regex, …), key-fn is applied to the matcher objects too, and a matcher can only be aligned to the value it matches by running it. For those cases, use in-any-order. sorted-by is intended for sequences 
of concrete, uniformly-typed, totally-orderable values.


test/clj/matcher_combinators/matchers_test.clj
```diff
@@ -81,6 +81,73 @@
...
+
+  (testing "catches a missing element (equals-like, not embeds-like)"
+    (is (no-match? (m/sorted-by :x [{:x 1} {:x 2} {:x 3}])
+                   [{:x 1} {:x 2}])))
+
+  (testing "non-sequential actual gives an invalid-type mismatch"
+    (is (match? {::result/type  :mismatch
+                 ::result/value mismatch?
+                 ::result/weight number?}
+                (c/match (m/sorted-by :x [{:x 1}]) {:x 1}))))
```
philomates: in running
```clojure
(is (match? (m/sorted-by :x [{:x 1}])
            {:x 3}))
```
it looks like the vector is converted to a list:

```clojure
;; eval: (is (match? (m/sorted-by :x [{:x 1}]) {:x 3}))
;; =>
;; FAIL in () (NO_SOURCE_FILE:38)
;; expected: (match? (m/sorted-by :x [{:x 1}]) {:x 3})
;;   actual: (mismatch (expected ({:x 1})) (actual {:x 3}))
```
so the `expected ({:x 1})` should actually be `expected [{:x 1}]`, similar to when you run 
```clojure
(is (match? [{:x 1}] {:x 3}))
```

test/clj/matcher_combinators/matchers_test.clj
```diff
@@ -81,6 +81,73 @@
...
+
+  (testing "composes with match-with: stays order-insensitive even though
+            match-with wraps the elements. This works because `sorted-by` sorts
+            `expected` eagerly at construction time, while it is still concrete
+            data, before match-with turns the elements into matchers. Both the
+            `expected` and the `actual` are out of order here."
+    (is (match? (m/match-with [map? m/equals]
+                              (m/sorted-by :x [{:x 2} {:x 1}]))
+                [{:x 1} {:x 2}])))
```
philomates: to properly show that it works with match-with, I think it would make sense to check a no-match when you `[map? m/equals]` and then have maps that embed but aren't actually equal.
for example:
```clojure
(is (match? (m/match-with [map? m/equals]
                          (m/sorted-by :x [{:x 2} {:x 1}]))
            [{:x 1} {:x 2 :y 'whatever}]))
```
similar to what we see here
```clojure
;; passes
(is (match? [{:x 2}]
            [{:x 2 :y 'whatever}]))
;; fails
(is (match? (m/match-with [map? m/equals]
                          [{:x 2}])
            [{:x 2 :y 'whatever}]))
```

philomates: Looks like my test suggestion passes when it should fail.
I think it would be important to get this working with `match-with` before mainlining it

test/clj/matcher_combinators/matchers_test.clj
```diff
@@ -81,6 +81,73 @@
...
+
+  (testing "LIMITATION: a bare matcher element has no concrete sort key, so
+            `sorted-by` cannot order it (sort-by throws, as matchers/predicates
+            are not Comparable). Use `in-any-order` for sequences of bare
+            matchers."
+    (is (thrown? Exception
+                 (m/sorted-by identity [odd? even?])))
```
philomates: The use of `sort key` here confuses me because it makes me think of `key-fn`, but I guess you are referring to something else. I would find it clearer to say that `sorted-by` doesn't work when the `key-fn` retrieves values that aren't sortable

src/cljc/matcher_combinators/matchers.cljc
```diff
@@ -105,6 +105,51 @@
...
+
+  LIMITATION: `key-fn` must be able to derive a sort key from each `expected`
+  element. This holds when elements are concrete values (including maps whose
+  inner values are submatchers, since the sort key is read from the concrete
+  outer structure). It does NOT hold when an element is *itself* a bare matcher
+  with no concrete sort key, e.g. `(sorted-by identity [odd? even?])`: a
+  predicate matches many values and has no single position to sort into, so
+  `sort-by` cannot order it (and in fact throws, as predicates are not
+  Comparable). For sequences of bare matchers, use `in-any-order`, which finds
+  the pairing by running the matchers."
```
philomates: I think this paragraph could be replaced with

> Limitation: values accessed under the key-fn must be sortable. Note that functions and matchers cannot be sorted. In such cases, consider the `in-any-order` matcher.

src/cljc/matcher_combinators/matchers.cljc
```diff
@@ -105,6 +105,51 @@
...
+
+   - Tied keys (non-injective `key-fn`): elements that share a key are left in
+     their incoming order, which may differ between `expected` and `actual`,
+     producing a spurious mismatch even though a valid pairing exists. Fix by
+     using a compound key that breaks ties, e.g. `(juxt :x :id)`.
+
+   - Mixed-type keys: `sort-by` throws (e.g. ClassCastException comparing a
+     String to a Long) — at construction time for `expected`, at match time for
+     `actual`. This surfaces as a thrown exception, not a mismatch, by design —
+     it signals misuse of `key-fn`.
```
philomates: hmm, I'm unsure if `ClassCastException`s are always misuses of `key-fn`/`sorted-by`.
For example:
you have a test in place that is passing, then the implementation changes and now you are seeing an exception. Does this mean you are now misusing the `sorted-by` matcher or is there a mismatch?

Regardless, I think it would be useful to see the data involved in the violation, be it data in the exception or data in the mismatch. Otherwise it could be hard to know what is causing it.

src/cljc/matcher_combinators/matchers.cljc
```diff
@@ -105,6 +105,51 @@
   Consider sorting the expected and actual sequences and comparing those instead."
```
philomates: the suggestion in `in-any-order` for dealing with poor performance could mention `sorted-by`

src/cljc/matcher_combinators/matchers.cljc
philomates: Another example regarding the `actual` not being sortable:
```clojure
(is (match? (m/sorted-by :a [{:a 1} {:a 2}])
            [{:a 2} {:a {:i-am-not-an-it 'foo}}]))
;; =>
;; actual: java.lang.ClassCastException: class clojure.lang.PersistentArrayMap cannot be cast to class java.lang.Comparable (clojure.lang.PersistentArrayMap is in unnamed module of loader 'app'; java.lang.Comparable is in module java.base of loader 'bootstrap')
```
Would it not make sense to catch the `ClassCastException` we get from trying to sort a list with an int and a map and instead return a mismatch.
I'm thinking this because imagine that the data was sorted already and there was no exception, like the following code snippet, we'd get a mismatch:
```clojure
(is (match? [{:a 1} {:a 2}]
            [{:a {:i-am-not-an-it 'foo}} {:a 2}]))
```

just thinking through the opposite, not sure if it makes sense yet

philomates (commented):
Hi Marco, thanks for the contribution and the new matcher seems quite useful!

I'm curious, is the `sorted-by` matcher something you've already started using in your day-to-day dev work? If so, how does it feel ergonomically? Like was it able to speed things up without much overhead to the dev experience?

marcojunior-code:
Hi @philomates!

Good question! Yeah, my team reaches for sorting whenever an array grows past a certain threshold (around 6 items, even though in-any-order can handle a bit more than that)

We don't have a shared helper for it yet, so the usual pattern is to define a sort function and sort both the expected and actual values in a binding before matching:

```
[let expected (sort-fn [...]
       actual      (sort-fn [...]
       (is (match? expected
                           actual))
```

The overhead is pretty low for us given we've settled on that threshold. We're just used to sorting once the array starts to grow. The whole idea behind sorted-by was essentially to push both sorts "under the hood" so we don't have to set up that binding by hand each time. This would also remove the need for a let binding in most cases.

(Thanks for the other comments by the way! I'm looking over them)

philomates:
@marcojunior-code thanks for the context and it sounds like the need comes up often enough that some sort of built-in would be worth adding.

I realized that maybe the bulk of what is achieved in the changes proposed here could be done using the `matcher-combinators.matchers/via` construct:

```clojure
(defn sorted-by [key-fn expected]
  (matcher-combinators.matchers/via 
   (partial sort-by key-fn)
   (sort-by key-fn expected)))
```
this seems to work with the problematic `match-with` example I gave before and in the case of a sorting exception it returns a mismatch due to a try/catch in the `via` impl.
```clojure
(is (match? (sorted-by :a [{:a 1} {:a 2}])
            [{:a 2} {:a {:i-am-not-an-it 'foo}}]))

; (out) FAIL in () (NO_SOURCE_FILE:18)
; (out) expected: (match? (sorted-by :a [{:a 1} {:a 2}]) [{:a 2} {:a {:i-am-not-an-it (quote foo)}}])
; (out)   actual: (mismatch
; (out)  (expected
; (out)   (via
; (out)         clojure.core$partial$fn__4759@3d84d2ac
; (out)         ({:a 1} {:a 2})))
; (out)  (actual [{:a 2} {:a {:i-am-not-an-it foo}}]))
```

I'll try to look a bit more at that possibility when I have some more time for OSS stuff in the next days

src/cljc/matcher_combinators/matchers.cljc
marcojunior-code: That's way better.

test/clj/matcher_combinators/matchers_test.clj
marcojunior-code: I noticed that this bug is fixable, but I really liked your idea of using the matchers/via. So in this case, I can add this test (which works with your solution), instead of fixing the current one.

src/cljc/matcher_combinators/matchers.cljc
marcojunior-code: Really nice catch, and I agree. It isn't always a misuse of key-fn. 
Using the "via" would probably solve this for the "actual", but we would still have problems with the "expected" I think.

marcojunior-code:
@philomates Seems like the "via" solution works way better. I tested it locally, but I made it that way:

```
(core/->ViaMatcher (fn [actual] (cond-> (sort-by key-fn actual)
                                          (vector? actual) vec))
                     (cond-> (sort-by key-fn expected)
                             (vector? expected) vec))
```

It can be more legible than that. I was just testing. 
The cond was added to prevent the return mentioned [here](https://github.com/nubank/matcher-combinators/pull/237#discussion_r3498933692)

Thanks for your time!

marcojunior-code:
Hi @philomates!
I saw your new contribution about reviewing LLM-generated codes.

I must say that in my case, the tests, comments, and docs were mostly generated by LLM, even though I was totally aware of the code and what it was supposed to do.

I really understand your position, and that's why I'm telling you that (but I think it was pretty obvious that part of the code was LLM-generated, given the amount of comments in it)

Anyway, I hope that we can still discuss this possible addition, but I will understand if you prefer not to.

philomates:
@marcojunior-code yeah, I'm still up for working on this PR. It is the first time I interact with code generated via LLM-tools, so I also wanted to look deeper out of curiosity :)

And going with `m/via` with the `vector?` condition sounds good.

The note I posted was mostly to clarify expectations going forward. It was instigated by your PR only in the sense that this is the first time I've knowingly interacted with LLM-generated code. So I hope it wasn't seen as a judgment of your work or even how you created it. My position on LLM-based tools is one that I established already months in the past: I believe that such tools can be used well and in a positive way, yet the current way they are trained, marketed, hosted, and distributed are so at odds with what I stand for, that I can't in good conscious support them in any way.
But this is my individual view and with it I'm not looking to create judgement or exclusion necessarily (though as a result there is unfortunately some exclusion..) but rather I want to exercise my ability to consciously choose how I interact with my surroundings.

marcojunior-code:
Thanks @philomates! Not feeling judged at all.

As a matter of fact, I really understand your point and your feelings about AI. So if I find an opportunity to make another contribution here, I will remember not to use LLM-generated code, since I noticed that your review was really on point and useful.

I will add the `m/via` and `vector?` solution into the branch probably after work today, and then I will ask for the review again.

Thanks again!

marcojunior-code:
Hi @philomates!
I updated the PR to use the ViaMatcher.
The problem of giving an exception when "actual" isn't sortable, and the composition of `match-with` was solved by this approach.

src/cljc/matcher_combinators/matchers.cljc
```diff
@@ -101,10 +101,29 @@
   `expected` list but with elements in a different order.
 
   WARNING: in-any-order can match each expected element against every value
-  in the actual sequence, which may be cost prohibitive for large sequences
-  Consider sorting the expected and actual sequences and comparing those instead."
+  in the actual sequence, which may be cost prohibitive for large sequences,
+  for these cases, consider using sorted-by approach instead"
```
philomates:
```suggestion
  for these cases, consider using the `sorted-by` matcher instead"
```

src/cljc/matcher_combinators/matchers.cljc
```diff
@@ -101,10 +101,29 @@
...
   [expected] (core/->InAnyOrder expected))
 
+(defn- sort-by-preserving-vector [key-fn coll]
+  (cond-> (sort-by key-fn coll)
+          (vector? coll) vec))
+
+(defn sorted-by
+  "Matcher that sorts both the `expected` and the `actual` sequences by
+  `key-fn` and then compares them in order (like `equals` over sequences).
```
philomates:
```suggestion
  `key-fn` and then invokes `match` on them
```

src/cljc/matcher_combinators/matchers.cljc
```diff
@@ -101,10 +101,29 @@
...
+
+  This is an O(n log n) alternative to `in-any-order` for large sequences:
```
philomates: The sorting is `O(n log n)` but I think the actual procedure may not be if the sorted inputs have `in-any-order`s nested in them. To avoid this ambiguity, I'd reword it to say "An alternative to `in-any-order` that avoids the exponential runtime that `in-any-order` incurs in its search for an minimal mismatch". Though also open to leaving it if you feel that having this measure is useful

src/cljc/matcher_combinators/matchers.cljc
```diff
@@ -101,10 +101,29 @@
...
+  rather than searching for a valid pairing between matchers and elements, it
+  relies on `key-fn` to line them up. Use it when you have a stable, cheap
+  sort key.
```
philomates: I would remove the note about "Use it when you have a stable, cheap sort key.".
In particular, I wonder: does the sort need to be stable to be useful? I suppose in cases where the resulting ordering isn't strictly monotonic you could get matches and mismatches in a nondeterministic fashion. But I think that is such a corner case that it isn't worth mentioning the need for stability. 
Though the more I think about it, isn't stability a function of the `clojure.core/sort-by` implementation and not having to do with the provided `key-fn`? Or am I confused?

And the user will probably understand that expensive sort key computations will lead to slower results.

src/cljc/matcher_combinators/matchers.cljc
```diff
@@ -101,10 +101,29 @@
...
+
+  Limitation: values accessed under the key-fn must be sortable. Note that
+  functions and matchers cannot be sorted. In such cases, consider the in-any-order matcher."
+  [key-fn expected]
+  (core/->ViaMatcher #(sort-by-preserving-vector key-fn %)
+                     (sort-by-preserving-vector key-fn expected)))
```
philomates: for a failure that results in a sorting exception
```clojure
(is (match? (m/sorted-by :a [{:a 1} {:a 2}])
            [{:a 2} {:a {:i-am-not-an-int 'foo}}]))
```
the output shows `via` instead of `sorted-by`:
```
 expected: (match? (m/sorted-by :a [{:a 1} {:a 2}]) [{:a 2} {:a {:i-am-not-an-int (quote foo)}}])
   actual: (mismatch
  (expected
   (via
         matcher_combinators.matchers$sorted_by$fn__18473@2efd53df
         [{:a 1} {:a 2}]))
  (actual [{:a 2} {:a {:i-am-not-an-int foo}}]))
```
I find this pretty confusing.
Since `->ViaMatcher` is an internal api only, we can tweak it a bit. So I think doing the following will give us useful mismatch output:
```clojure
(defrecord ViaMatcher [base-name transform-actual-fn expected]
  Matcher
  (-matcher-for [_this] (-matcher-for expected))
  (-matcher-for [_this x] (-matcher-for expected x))
  (-match [this actual]
    (let [transformed (try (transform-actual-fn actual)
                           (catch #?(:clj Exception :cljs js/Error) e e))]
      (if (instance? #?(:clj Exception :cljs js/Error) transformed)
        {::result/type   :mismatch
         ::result/value  (model/->Mismatch
                          (list (-base-name this) expected)
                          actual)
         ::result/weight 1}
        (match expected transformed))))
  (-base-name [_] base-name))
```
and then `sorted-by` body becomes
```clojure
(defn sorted-by
  [key-fn expected]
  (core/->ViaMatcher (symbol (str "sorted-by " key-fn))
                     #(sort-by-preserving-vector key-fn %)
                     (sort-by-preserving-vector key-fn expected)))
```
and we get
```
expected: (match? (m/sorted-by :a [{:a 1} {:a 2}]) [{:a 2} {:a {:i-am-not-an-int (quote foo)}}])
  actual: (mismatch
 (expected (sorted-by :a [{:a 1} {:a 2}]))
 (actual [{:a 2} {:a {:i-am-not-an-int foo}}]))
```

(`via` needs to be updated in a similar manner)

After considering these changes I now feel strange about the fact that the exception that happens is completely hidden to the user. Ideally we'd expose that somehow but I think it is work for another chunk of work because it also touches the UX of `via`.

test/clj/matcher_combinators/matchers_test.clj
```diff
@@ -81,6 +81,45 @@
                        ::result/value
                        (map vals)))))))
 
+(deftest sorted-by-matcher
```
philomates, line 84: We should have tests showing the behavior of `sorted-by` when the `actual` or the `expected` are sets

I think things are mostly fine, but there is a mismatch in the sequence type in the mismatch message for things like
```clojure
(is (match? (m/sorted-by :a #{{:a 1}})
            #{{:a 1} {:a 2 :b 2}}))
```
where the `actual` input to `sorted-by` is a set but in the output it shows as a list.
```
expected: (match? (m/sorted-by :a #{{:a 1}}) #{{:a 1} {:a 2, :b 2}})
  actual: ({:a 1} (unexpected {:a 2, :b 2}))
```
I guess this is fine. We could alternatively look into casting it back to a set but the ordering of elements isn't guaranteed, so the output might be in a different order than what `sorted-by` produced. Maybe an ordered set, but I don't think it is worth the trouble

src/cljc/matcher_combinators/matchers.cljc
```diff
@@ -101,10 +101,29 @@
...
+  "Matcher that sorts both the `expected` and the `actual` sequences by
+  `key-fn` and then compares them in order (like `equals` over sequences).
+
+  This is an O(n log n) alternative to `in-any-order` for large sequences:
+  rather than searching for a valid pairing between matchers and elements, it
+  relies on `key-fn` to line them up. Use it when you have a stable, cheap
+  sort key.
+
+  Limitation: values accessed under the key-fn must be sortable. Note that
+  functions and matchers cannot be sorted. In such cases, consider the in-any-order matcher."
```
philomates: the other limitation is that in cases where the result of the sorting isn't strictly monotonic, that is, there is an equivalence between two results of the key-fn, `sorted-by` may produce mismatches where `in-any-order` matches. e.g. 
```clojure
(is (match? (m/sorted-by :x [{:x 1 :y odd?} {:x 1 :y even?}])
            [{:x 1 :y 2} {:x 1 :y 1}]))
```

README.md
```diff
@@ -175,6 +175,8 @@ for a specific value, e.g.
 - `in-any-order` operates over sequences
 
   matches when the given a sequence that is the same as the `expected` sequence but with elements in a different order.  Similar to midje's `(just expected :in-any-order)`
+- `sorted-by` takes a `key-fn` and an `expected` sequence, sorts both `expected` and `actual` by `key-fn`, and matches them in order. 
+   An alternative to `in-any-order` for large sequences of sortable values.
```
philomates:
```suggestion
   An alternative to `in-any-order` for cases when inputs are sortable and `in-any-order` has poor performance, such as with large sequences.
```

test/clj/matcher_combinators/matchers_test.clj
```diff
@@ -81,6 +81,45 @@
...
+    (is (no-match? (m/sorted-by :x [{:x 1} {:x 2} {:x 3}])
+                   [{:x 1} {:x 2}])))
+
+  (testing "non-sequential actual doesn't match"
+    (is (no-match? (m/sorted-by :x [{:x 1}]) {:x 1})))
+
+  (testing "match-with actually applies the given matcher"
+    (is (no-match? (m/match-with [map? m/equals]
+                              [{:x 2}])
+                [{:x 2 :y 'whatever}])))
```
philomates: this test doesn't include `sorted-by`

test/clj/matcher_combinators/matchers_test.clj
```diff
@@ -81,6 +81,45 @@
...
+
+  (testing "functions and matchers aren't sortable, but values accessed under the key-fn must be.
+            Use `in-any-order` for sequences of bare matchers."
+    (is (thrown? Exception
+                 (m/sorted-by identity [odd? even?])))
```
philomates: we also want to test when the `actual` isn't sortable but the the `expected` is. Like showing clearly the difference between the two: how we get a mismatch instead of an exception.

src/cljc/matcher_combinators/matchers.cljc
marcojunior-code: Makes sense. I forgot about the possibility of having an in-any-order inside of it.

src/cljc/matcher_combinators/matchers.cljc
marcojunior-code: Yeah, thinking more about it, maybe we only need to inform that the key-fn must be sortable. We can leave it to the user to decide whether or not to use it when it comes to expensive computations

test/clj/matcher_combinators/matchers_test.clj
marcojunior-code: I will add tests to document the scenario, but I agree that casting it back to a set is not worth it right now.

src/cljc/matcher_combinators/matchers.cljc
marcojunior-code:
```
Limitation: When two elements yield the same `key-fn` value, sorted-by can report a mismatch
  even when a valid pairing exists.
  e.g. `(sorted-by :x [{:x 1 :y odd?} {:x 1 :y even?}])` fails against `[{:x 1 :y 2} {:x 1 :y 1}]`.
  Prefer in-any-order when keys aren't unique, and elements are matched with submatchers.
```

I will add this limitation. We also have a test to document it.

test/clj/matcher_combinators/matchers_test.clj
marcojunior-code: I added it and also changed the test description.

src/cljc/matcher_combinators/matchers.cljc
marcojunior-code: I made the update here. Do you think we need a test to validate the mismatch for the via? 
Something like:
```
  (testing "erroring shows `(mismatch (expected (via some-fn expected-data)) ...)`"                                                                   
    (let [value (::result/value                                                                                                                       
                 (c/match {:payloads [(m/via read-string {:foo :barz})]}                                                                              
                          {:payloads [1]}))                                                                                                           
          via-mismatch (first (:payloads value))]                                                                                                     
      (is (mismatch? via-mismatch))                                                                                                                   
      (is (= 'via (first (:expected via-mismatch))))))
```

The payload should be:
`{:payloads [{:expected (via {:foo :barz}), :actual 1}]}`

marcojunior-code: > Ideally we'd expose that somehow but I think it is work for another chunk of work because it also touches the UX of via.

And yes, for me it doesn't feel right to do this in this PR.

philomates: okay, once my suggested changes to `ViaMatcher` get added allowing for configuring the base-name, we can wrap this up.
a test like you suggested would be good probably, though more important is that we test that `sorted-by` doesn't show `via` in the `:expected`, since that is the behavior we want to avoid but is present in the current implementation

philomates:
forgot to follow up, but this change is deployed in `3.11.0` last week :tada:

marcojunior-code:
Hey @philomates! Thanks!
Sorry for kind of disappearing. I was on vacations and another feature is taking most of my time here.

